PR #68, opened 2023-01-03
Bump ameba + fix issues found by the upcoming ameba version (1.4.0)

Sija:
Hi @ellmetha,

First of all, thanks for this new 👶 on the block! :)

As an ameba contributor I wanted to ensure you're having a smooth ride with both Crystal, and Ameba, and so I took a liberty to bump it to the last version, clean up the configuration and fix several issues that popped up.

I hope you'll find it helpful.

Best,
Sija

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -1,117 +1,17 @@
-Globs:
-  - spec/**/*.cr
-  - src/**/*.cr
-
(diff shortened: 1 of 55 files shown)


ellmetha (approved):
Thanks a lot for this @Sija!

Issue #64, opened 2022-12-21
Add support for linebreaks template filter


ellmetha:
# Description

Marten's templating system should provide a convenient `linebreaks` template filter allowing to replace all newlines in a text to HTML line breaks (`<br />`).

For example, consider the following snippet:

```html
{{ value | linebreaks }}
```

If `value` is `Hello\nWorld`, then the output of the rendered snippet would be `Hello<br />World`.

ellmetha:
Added in bff795f!

PR #69, opened 2023-01-07
Linebreaks

dredge:
Pull request to address **Add support for linebreaks template filter #64**

My first ever.

Crossing fingers.

-Chris


src/marten/template/filter/linebreaks.cr
```diff
@@ -0,0 +1,14 @@
+module Marten
+  module Template
+    module Filter
+      # The `linebreaks` filter.
+      #
+      # The `linebreaks` filter allows to convert a string replacing all newlines with HTML line breaks (<br />).
+      class LineBreaks < Base
+        def apply(value : Value, arg : Value? = nil) : Value
+          Value.from(value.to_s.gsub("\n", "<br />"))
```
ellmetha, line 9: In Marten templates, `String` objects are automatically [HTML-escaped](https://martenframework.com/docs/templates/introduction#auto-escaping). This means that with the present implementation, the `<br />` tags injected by this filter would be escaped by the Marten template engine automatically when rendering templates (eg. to `&lt;br /&gt;`).

To avoid this, you will need to convert the string to a "safe string" object in order to tell Marten to not automatically escape the string returned by the `linebreaks` filter. But by doing this we would also make the use of `linebreaks` vulnerable to Cross-Site Scripting (XSS) attacks because the content of `value` would be marked as "safe" and rendered as HTML as well. So we need to do two things:

1. HTML-escape the incoming `value` and replace the `\n` characters
2. Mark the resulting string as "safe"

This could be achieved with something like that:

```suggestion
          Value.from(SafeString.new(HTML.escape(value.to_s).gsub("\n", "<br />")))
```

Obviously, this will also involve some changes at the specs level! 🙂

ellmetha (commented):
Thanks a lot for looking into that! 😃 

The code looks good, there is just a small issue related to safe strings and auto-escaping that needs to be addressed before merging. See my comment below!

dredge:
Thank you for the helpful input - this should be closer to what is needed.
-Chris

ellmetha (approved):
LGTM! Thanks a lot @dredge! 😃

PR #77, opened 2023-02-20
Remove duplicated fields reference: auto_now and auto_now_add

vmrocha:
Remove duplicated reference documentation for fields `auto_now` and `auto_now_add`.


vmrocha:
My bad, I just understood now that these are arguments for both `date` and `date_time`. 👍

Issue #63, opened 2022-12-12
Add support for split and join template filters


ellmetha:
# Description

It should be possible to rely on `split` and `join` filters in templates. The envisioned syntax for these filters is highlighted below:

```
{% assign fruits = "Banana,Orange,Apple" | split: "," %}
{{ fruits | join: ", " }}
```

ellmetha:
Fixed in f0d70d2.

PR #78, opened 2023-03-05
Add support for split and join template filters #63

usiegj00:
I've added two new filters, filter specs and additions to docs/filters to add the `join` and `split` filter functionality. My first contribution here, so please give brutal feedback (with a crystal, marten or general contributor lens).

### DIFF
--- a/src/marten/template/filter/join.cr
+++ b/src/marten/template/filter/join.cr
@@ -3,11 +3,13 @@ module Marten
     module Filter
       # The "join" filter.
       #
-      # The "join" filter converts an array into a string with elements separated by `arg`.
+      # The "join" filter converts an enumerable into a string with elements separated by `arg`.
       class Join < Base
         def apply(value : Value, arg : Value? = nil) : Value
-          # Only work on array
-          return value if !value.raw.is_a?(Array)
+          # Only work on enumerable
+          if !value.raw.is_a?(Enumerable)
+            raise Errors::UnsupportedType.new("#{value.raw.class} objects can't be joined")
+          end
           if arg
             Value.from(value.to_a.map(&.to_s).join(arg.to_s))
           else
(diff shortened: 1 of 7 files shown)


src/marten/template/filter/join.cr
```diff
@@ -0,0 +1,20 @@
+module Marten
+  module Template
+    module Filter
+      # The "join" filter.
+      #
+      # The "join" filter converts an array into a string with elements separated by `arg`.
+      class Join < Base
+        def apply(value : Value, arg : Value? = nil) : Value
+          # Only work on array
+          return value if !value.raw.is_a?(Array)
```
ellmetha, line 10: I think we'll want to ensure that the `join` filter can be used for all [`Enumerable`](https://crystal-lang.org/api/1.7.2/Enumerable.html) objects (because all of these could respond to `#to_a`). Moreover, for filters that work only for specific types of values, we usually raise an error to indicate that the filter was misused!

For example:

```crystal
def apply(value : Value, arg : Value? = nil) : Value
  if !value.raw.is_a?(Enumerable)
    raise Errors::UnsupportedType.new("#{value.raw.class} objects can't be joined")
  end

  if arg
    Value.from(value.to_a.map(&.to_s).join(arg.to_s))
  else
    Value.from(value.to_a.map(&.to_s).join)
  end
end
```

ellmetha (commented):
Thanks for taking care of this @usiegj00! 🙏 

Just one thing to fix before merging (see my comment below). Otherwise, we should also add a few specs for these new filters ideally to test them properly! 🙂

src/marten/template/filter/join.cr
usiegj00: I think I've done this!

ellmetha (approved):
LGTM! Thanks a lot for this @usiegj00! 😃

Issue #75, opened 2023-02-15
Add the ability to check if a query set contains a specific record or not


ellmetha:
# Description

Presently, it is possible to verify if a record is part of a query set by using the standard `#includes?` method. This works because query sets are enumerable objects, but the drawback is that doing so results in all the records of the query sets being retrieved and stored in memory (which can be a problem for large query sets).

It would be interesting to be able to perform membership checks like that without fetching all the records of the query set. In this light, let's add the ability to perform these more efficient checks by overriding the standard `#includes?` method so that it performs a dedicated SQL query in order to determine if the record is part of the query set or not.

PR #80, opened 2023-03-07
Add #includes? to Set.

usiegl00:
See #75

### DIFF
--- a/src/marten/db/query/set.cr
+++ b/src/marten/db/query/set.cr
@@ -312,7 +312,7 @@ module Marten
         #
         # ```
         # query_set = Post.all
-        # query_set.filter { (q(name: "Foo") | q(name: "Bar")) & q(is_published: True) }
+        # query_set.filter { (q(name: "Foo") | q(name: "Bar")) & q(is_published: true) }
         # ```
         def filter(&)
           expr = Expression::Filter.new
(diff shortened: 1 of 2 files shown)


src/marten/db/query/set.cr
```diff
@@ -325,6 +325,40 @@ module Marten
...
+
+        # Returns if a query set includes a specific node.
+        # ```
+        # query_set = Post.all
+        # query_set.includes?(Marten::DB::Query::Node.new(name__startswith: "Fr"))
+        # ```
+        def includes?(query_node : Node)
+          filter(query_node)
+          exists?
+        end
```
ellmetha, line 360: I am not opposed to adding these methods. Though in my mind they should be called `#exists?` since what we are doing is verifying that the query set produces results (which would also be more consistent with the existing `#exists?` method). Can we rename those to `#exists?`?

src/marten/db/query/set.cr
```diff
@@ -325,6 +325,40 @@ module Marten
           add_query_node(query_node)
         end
 
+        # Returns if a query set includes a specific model.
+        def includes?(value : DB::Model)
```
ellmetha, line 329: Can we use `M` instead of `DB::Model`? The `Marten::DB::Query::Set` abstraction is a [generic class](https://github.com/martenframework/marten/blob/f0d70d278f814c56777fac7c1317426cc8a8e707/src/marten/db/query/set.cr#L14).

```suggestion
        def includes?(value : M)
```

This will ensure that only `Article` records can be used for `Marten::DB::Query::Set(Article)` queru sets objects for example.

src/marten/db/query/set.cr
```diff
@@ -325,6 +325,40 @@ module Marten
...
+          query_node = Node.new({Constants::PRIMARY_KEY_ALIAS => value.pk})
+          add_query_node(query_node)
+          exists?
```
ellmetha, line 332: Note that `#add_query_node` returns a _new_ query set object (query sets are _never_ mutated). So presently, the `#exists?` call is executed on the initial (and unfiltered) query set.

Another thing to take into account is that the query set could already be evaluated when this method is executed: in this situation, we can avoid hitting the DB again by simply calling `#includes?` on the cached array of records.

To fix all this, you could simply do something like this:

```suggestion
          if @result_cache.nil?
            filter(Node.new({Constants::PRIMARY_KEY_ALIAS => value.pk})).exists?
          else
             @result_cache.not_nil!.includes?(value)
           end
```

src/marten/db/query/set.cr
ellmetha, line 329: Another thing to take into account is that we probably want to check that the specified record is persisted. If that's not the case and if the record does not have a primary key yet, we could raise a `Marten::DB::Errors::UnmetQuerySetCondition` exception explaining the situation.

What do you think?

ellmetha (commented):
Thanks for taking a stab at this @usiegl00!

That's definitely a good start, and I left a few comments below. I think we will also need to update the documentation. 📝

src/marten/db/query/set.cr
ellmetha: You can simply rename these methods to `#exists?` because afaik `#includes?` has only [one interface](https://crystal-lang.org/api/1.7.2/Enumerable.html#includes%3F%28obj%29%3ABool-instance-method) where you specify one object as argument (ie. the object to use for the inclusion check).

src/marten/db/query/set.cr
ellmetha: There is a [`#persisted?`](https://martenframework.com/docs/api/0.2/Marten/DB/Model/Persistence.html#persisted%3F-instance-method) method available on all model record objects that you can use!

src/marten/db/query/set.cr
```diff
@@ -325,6 +325,41 @@ module Marten
           add_query_node(query_node)
         end
 
+        # Returns if a specific model exists in the query set.
+        def exists?(value : M)
```
ellmetha, line 329: I think the name of this specific method only could've remained `#includes?` in order to fix the initial issue. I'll fix this after merging this.

ellmetha (approved):
Thanks for taking care of this. This is a nice improvement! 👍

PR #105, opened 2023-06-23
Implement with tag for template

### LINKED ISSUE #86: Add a 'with' template tag
# Description

Presently, new template variables can be easily assigned from within a template by leveraging the [`assign`](https://martenframework.com/docs/templates/reference/tags#assign) template tag. This works fine, but there is presently no way to "locally" define a variable that shouldn't be available to the whole template.

To palliate that, let's introduce a `with` tag that would work like that:

```
{% with val1="foo", val2="bar" %}
  {{ val1 }}
  {{ val2 }}
{% endwith %}
```

The defined variables should only be available within the `{% with %}...{% endwith %}` tags, which could easily be achieved by leveraging the existing [template context stacking capability](https://github.com/martenframework/marten/blob/2213da45cedc409b17ddbaa19e6351445917fa93/src/marten/template/context.cr#L88).

treagod:
I've added a `with` tag to assign variables locally within the template.

I'm not quite sure about the last test case if the order should matters.  Maybe it would be better to first look in the context if any variable name is already defined and shadow the outer variable only after assignment was done.

Closes #86

### DIFF
--- a/spec/marten/template/tag/with_spec.cr
+++ b/spec/marten/template/tag/with_spec.cr
@@ -1,0 +1,129 @@
+require "./spec_helper"
+
+describe Marten::Template::Tag::With do
+  describe "::new" do
+    it "raises if the with block does not contain at least one assignment" do
+      parser = Marten::Template::Parser.new(
+        "test{% endwith %}"
+
+      )
+
+      expect_raises(
+        Marten::Template::Errors::InvalidSyntax,
+        "Malformed with tag:at least on assignment must be present"
+      ) do
+        Marten::Template::Tag::With.new(parser, "with x")
+      end
+    end
+  end
+
+    describe "#render" do
+      it "properly renders a simple with block" do
+        parser = Marten::Template::Parser.new(
+          <<-TEMPLATE
+          {{ x }}
+          {% endwith %}
+          TEMPLATE
+        )
+        tag = Marten::Template::Tag::With.new(parser, "with x=1")
+
+        tag.render(Marten::Template::Context.new).strip.should eq "1"
+      end
+    end
+
+      it "properly renders with multiple variables assigned" do
+        parser = Marten::Template::Parser.new(
+          <<-TEMPLATE
+          {{ x }} - {{ y }}
+          {% endwith %}
+          TEMPLATE
+        )
+        tag = Marten::Template::Tag::With.new(parser, "with x=1, y=2")
+
+        tag.render(Marten::Template::Context.new).strip.should eq "1 - 2"
+      end
+
+      it "raises if a variable was defined but not set" do
+        parser = Marten::Template::Parser.new(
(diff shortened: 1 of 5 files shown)


src/marten/template/tag/with.cr
```diff
@@ -0,0 +1,49 @@
...
+              "Malformed with tag:at least on assignment must be present"
+            )
+          end
+        end
+
+        def render(context : Context) : String
+          context.stack do |with_context|
+            String.build do |io|
+              @assignments.each do |assignment|
+                context[assignment[0]] = FilterExpression.new(assignment.[1]).resolve(context)
```
ellmetha, line 39: Ideally the `FilterExpression` should be initialized in the `#initialize` method and then resolved in the `#render` method. Indeed, `FilterExpression#new` will actually parse the raw filter expression (eg. `myvar|upcase`) and ensure that it is valid from a template syntax standpoint. If we do that in the `#render` method, this means that we will parse each filter expression every time the template is rendered!

One idea: maybe the `@assignments` instance variable could be an array of `Tuple(String, FilterExpression)` instead? This means that you would need to probably do something like that in the `#initialize` method in order to initialize the `@assignments` variable accordingly:

```crystal
def initialize(parser : Parser, source : String)
  assignments = extract_assignments(source)
  @with_nodes = parser.parse(up_to: {"endwith"})
  parser.shift_token

  if assignments.size == 0
    raise Errors::InvalidSyntax.new(
      "Malformed with tag:at least on assignment must be present"
    )
  end

  @assignemnts = assignments.map do |name, raw_filter_expression|
    {name, FilterExpression.new(raw_filter_expression)}
  end
end
```

What do  you think?

spec/marten/template/tag/with_spec.cr
```diff
@@ -0,0 +1,127 @@
...
+
+    expect_raises(
+      Marten::Template::Errors::UnknownVariable,
+      "Failed lookup for attribute 'y'"
+    ) do
+      tag.render(Marten::Template::Context.new)
+    end
+  end
+
+  it "context for the variables is locally" do
```
ellmetha, line 62:
```suggestion
  it "does not pollute the outer context with local variables" do
```

ellmetha (commented):
Thanks a lot for taking care of this! 🙏 

There is just a minor improvements to make in order to ensure that the `with` filter expressions are actually parsed at parsing time (and not at rendering time), and then we should be able to merge this. See my comment below! 🙂

src/marten/template/tag/with.cr
treagod: Sounds good and reasonable 😄 thanks for having a look at it.

I will update the commit later that day

ellmetha (approved):
LGTM! Thanks a lot for adding this! 🚀

PR #106, opened 2023-06-24
Add --database to new command

### LINKED ISSUE #103: Add the ability to generate projects with a specific database backend
# Description

New projects generated via the use of the [`new`](https://martenframework.com/docs/development/reference/management-commands#new) automatically use a local SQLite database. It would be interesting to be able to generate new projects that make use of a PostgreSQL or MySQL database too. In this light, let's add a `--database` option whose value could be either `sqlite`, `mysql`, or `postgresql`.

```bash
marten new project test --database=postgresql
marten new project test --database=mysql
marten new project test --database=sqlite
```

If not specified, the default value should be `sqlite` (which corresponds to the existing behavior).

When used, this option should result in the right database backend being configured in the generated project's settings and the corresponding shard ([crystal-pg](https://github.com/will/crystal-pg), [crystal-mysql](https://github.com/crystal-lang/crystal-mysql), or [crystal-sqlite3](https://github.com/crystal-lang/crystal-sqlite3)) added to the project's `shard.yml` file.

treagod:
This commit adds the --database option to the new command.

By default when the option is not presented the sqlite3 database will be selected, otherwise the respective database which was choosen, i.e. the shard.yml file will be configured to include the shard for the specific database.

If MySql or Postgresql is selected the config/base.cr will use `localhost` as host and `<project_name>_db` as the database name. The user and password config will be written as a comment to `config/base.cr` as we can't make assumptions.

Closes #103

### DIFF
--- a/docs/docs/development/reference/management-commands.md
+++ b/docs/docs/development/reference/management-commands.md
@@ -113,6 +113,7 @@ The command allows you to fully define the name of your project or application, 
 
 * `-d DIR, --dir=DIR` - An optional destination directory
 * `--with-auth` - Adds an authentication application to newly created projects. See [Authentication](../../authentication) to learn more about this capability
+* `--database` - Configure database. Currently MySql, Postgresql and Sqlite3 are supported. See [Database settings](../settings#database-settings) for more information.
 
 ### Arguments
 
--- a/docs/docs/development/reference/management-commands.md
+++ b/docs/docs/development/reference/management-commands.md
@@ -113,7 +113,7 @@ The command allows you to fully define the name of your project or application, 
 
 * `-d DIR, --dir=DIR` - An optional destination directory
 * `--with-auth` - Adds an authentication application to newly created projects. See [Authentication](../../authentication) to learn more about this capability
-* `--database` - Configure database. Currently MySql, Postgresql and Sqlite3 are supported. See [Database settings](../settings#database-settings) for more information.
+* `--database` - Configures the applications database. Currently MySql, Postgresql and Sqlite3 are supported. See [Database settings](../settings#database-settings) for more information.
 
 ### Arguments
 
--- a/docs/docs/development/reference/management-commands.md
+++ b/docs/docs/development/reference/management-commands.md
@@ -113,7 +113,7 @@ The command allows you to fully define the name of your project or application, 
 
 * `-d DIR, --dir=DIR` - An optional destination directory
 * `--with-auth` - Adds an authentication application to newly created projects. See [Authentication](../../authentication) to learn more about this capability
-* `--database` - Configures the applications database. Currently MySql, Postgresql and Sqlite3 are supported. See [Database settings](../settings#database-settings) for more information.
+* `--database` - Configures the applications database. Currently `mysql`, `postgresq` and `sqlite3` are supported. See [Database settings](../settings#database-settings) for more information.
 
 ### Arguments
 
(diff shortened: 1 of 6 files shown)


src/marten/cli/manage/command/new.cr
```diff
@@ -30,6 +31,12 @@ module Marten
             on_argument(:name, "Name of the project or app to initialize") { |v| @name = v }
             on_option_with_arg(:d, :dir, arg: "dir", description: "Optional destination directory") { |v| @dir = v }
             on_option(:"with-auth", "Adds authentication to newly created projects") { @with_auth = true }
+            on_option_with_arg(
+              :database,
```
ellmetha, line 35: Can we also provide a short flag alternative to this new option (eg. `-db`)? 

```suggestion
            on_option_with_arg(
              :db,
              :database,
```

src/marten/cli/manage/command/new/templates/project/config/settings/base.cr.ecr
```diff
@@ -26,10 +26,28 @@ Marten.configure do |config|
...
     db.backend = :sqlite
     db.name = Path["<%= @context.name %>.db"].expand
   end
+  <%- elsif @context.database == "mysql" -%>
+  config.database do |db|
+    db.backend = :mysql
+    db.host = "localhost"
+    db.name = "<%= @context.name %>_db"
+    # db.user = "my_user"
+    # db.password = "my_passport"
```
ellmetha, line 40:
```suggestion
    # db.password = "my_password"
```

src/marten/cli/manage/command/new/templates/project/config/settings/base.cr.ecr
```diff
@@ -26,10 +26,28 @@ Marten.configure do |config|
...
+  end
+  <%- elsif @context.database == "postgresql" -%>
+  config.database do |db|
+    db.backend = :postgresql
+    db.host = "localhost"
+    db.name = "<%= @context.name %>_db"
+    # db.user = "my_user"
+    # db.password = "my_passport"
```
ellmetha, line 48:
```suggestion
    # db.password = "my_password"
```

docs/docs/development/reference/management-commands.md
```diff
@@ -113,6 +113,7 @@ The command allows you to fully define the name of your project or application,
 
 * `-d DIR, --dir=DIR` - An optional destination directory
 * `--with-auth` - Adds an authentication application to newly created projects. See [Authentication](../../authentication) to learn more about this capability
+* `--database` - Configures the applications database. Currently MySql, Postgresql and Sqlite3 are supported. See [Database settings](../settings#database-settings) for more information.
```
ellmetha, line 116: I think we can also explicitly mention the expected raw values here (`mysql/postgresql/sqlite3`).

ellmetha (commented):
Thanks for taking care of this! 🙏 

This looks good to me! There are just a few minor details to improve before merging (see my comments below) but all good otherwise! 👍

src/marten/cli/manage/command/new.cr
treagod: I've tried it, but apparently the crystal `OptionParser` allows only one character for the short flag. 

If `:db` is defined as a short flag and a new project is created with `marten new -db=mysql project dummy_project` then a project `b=mysql` will be created (because `-d` is used with the value `b=mysql`)

ellmetha: Sounds good. Let's keep it that way!

ellmetha (approved):
LGTM! 👍

PR #107, opened 2023-06-24
Add command to remove expired sessions

treagod:
Adds a new abstract method `clear_expired_entries` to `Marten::HTTP::Session::Store::Base`.

Additionally a new CLI command clearsessions was added, which calls `clear_expired_entries` on all registered session stores

### DIFF
--- a/src/marten/cli/manage/command/clearsessions.cr
+++ b/src/marten/cli/manage/command/clearsessions.cr
@@ -5,15 +5,16 @@ module Marten
         class ClearSessions < Base
           command_name :clearsessions
           help "Clears all expired sessions."
+          getter no_input
 
-          @no_prompt : Bool = false
+          @no_input : Bool = false
 
           def setup
-            on_option("-y", "Do not show prompts to the user") { @no_prompt = true }
+            on_option("no-input", "Do not show prompts to the user") { @no_input = true }
           end
 
           def run
-            unless no_prompt?
+            unless no_input
               print("All expired sessions will be removed.")
               print("These sessions can't be restored.")
               print("Do you want to continue [yes/no]?", ending: " ")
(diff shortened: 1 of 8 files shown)


src/marten/cli/manage/command/clearsessions.cr
```diff
@@ -0,0 +1,41 @@
...
+    class Manage
+      module Command
+        class ClearSessions < Base
+          command_name :clearsessions
+          help "Clears all expired sessions."
+
+          @no_prompt : Bool = false
+
+          def setup
+            on_option("-y", "Do not show prompts to the user") { @no_prompt = true }
```
ellmetha, line 12: Can we use the `no-input` instead of `-y` to be consistent with similar options in other commands (for example like in the [`collectassets`](https://github.com/martenframework/marten/blob/3e7365566e7c955d8c927d29ee0fbf74d5dc2a4c/src/marten/cli/manage/command/collect_assets.cr#L12) command)?

```suggestion
            on_option("no-input", "Do not show prompts to the user") { @no_input = true }
```

src/marten/cli/manage/command/clearsessions.cr
```diff
@@ -0,0 +1,41 @@
...
+                return
+              end
+            end
+
+            print(style("Clearing expired sessions", fore: :light_blue, mode: :bold), ending: "\n")
+
+            HTTP::Session::Store.registry.each_value do |session_store_klass|
+              store = session_store_klass.new(nil)
+              store.clear_expired_entries
+            end
```
ellmetha, line 31: Technically we should only clear expired entries for the configured sessions store. For example you could ensure this is the case by doing something like that:

```suggestion
            session_store_klass = HTTP::Session::Store.get(Marten.settings.sessions.store)
            session_store_klass.new(nil).clear_expired_entries
```

That `#clear_expired_entries` will only get called on the cookie store if this is the one being used by the current Marten project. 🙂

src/marten/cli/manage/command/clearsessions.cr
```diff
@@ -0,0 +1,41 @@
...
+          end
+
+          private def no_prompt?
+            @no_prompt
+          end
```
ellmetha, line 36: You can simplify this by using `#getter?`:

```suggestion
          private getter? no_prompt
```

ellmetha (commented):
Hey @treagod! Thanks for submitting this! 🙏 

This looks good overall and there will be just some minor improvements to make to the PR before merging it! We should also document the new management command ideally, by mentioning it in the [`command reference`](https://github.com/martenframework/marten/blob/main/docs/docs/development/reference/management-commands.md). 🙂

ellmetha (approved):
Thanks for adding this! 👍

PR #114, opened 2023-07-30
Make name parameter optional for included path

### LINKED ISSUE #108: Make it possible to include route maps without specifying a namespace
# Description

It is presently possible to define included routes by leveraging the `Marten::Routing::Map#draw` method and by defining a corresponding "path":

```crystal
ARTICLE_ROUTES = Marten::Routing::Map.draw do
  path "", ArticlesHandler, name: "list"
  path "/create", ArticleCreateHandler, name: "create"
  path "/<pk:int>", ArticleDetailHandler, name: "detail"
  path "/<pk:int>/update", ArticleUpdateHandler, name: "update"
  path "/<pk:int>/delete", ArticleDeleteHandler, name: "delete"
end

Marten.routes.draw do
  path "/", HomeHandler, name: "home"
  path "/articles", Articles::ROUTES, name: "articles"
end
```

Defining nested routes like that always requires the definition of an associated `name` for the including route. While this may work fine in some situations, it should be possible to define included routes without having to define a namespace at the including route level (since all the underlying routes must have an associated name):

```crystal
ARTICLE_ROUTES = Marten::Routing::Map.draw do
  path "", ArticlesHandler, name: "article_list"
  path "/create", ArticleCreateHandler, name: "article_create"
  path "/<pk:int>", ArticleDetailHandler, name: "article_detail"
  path "/<pk:int>/update", ArticleUpdateHandler, name: "article_update"
  path "/<pk:int>/delete", ArticleDeleteHandler, name: "article_delete"
end

Marten.routes.draw do
  path "/", HomeHandler, name: "home"
  path "/articles", Articles::ROUTES
end
```

Let's ensure that the use of the `name` argument for including routes is optional when calling the `#path` method. The use of the `name` option should remain mandatory when defining routes for handlers.

treagod:
This commit makes the `name` parameter for included maps optional. In addition a namespace can be added to a map. 

Example

```crystal
ARTICLE_ROUTES = Marten::Routing::Map.draw(:article) do # :article is the namespace
  path "/", ArticleList, name: "list"
  path "/new", ArticleNew, name: "new"
end

Marten.routes.draw do
  path "/articles", ARTICLE_ROUTES # will add article:list and article:new
end
```

Closes #108


treagod:
I don't know why, but shards install raises a compiler bug. Seems to be linked to ameba?

src/marten/routing/map.cr
```diff
@@ -20,13 +20,25 @@ module Marten
...
-      # or another `Marten::Routing::Map` instance (in case of nested routes maps). Each <path, target> pair must be
-      # given a name that will be used to uniquely identify the route.
-      def path(path : String, target : Marten::Handlers::Base.class | Map, name : String | Symbol) : Nil
-        if name.empty?
-          raise Errors::InvalidRuleName.new("Route names cannot be empty")
-        end
+      # The target associated with the considered path must be a `Marten::Routing::Map`
+      # instance. The <path, target> pair has an optional
+      # name that will be prepended to each <path, target> pair inside the `Marten::Routing::Map`.
+      def path(path : String, target : Map, name : String | Symbol = "") : Nil
```
ellmetha, line 26: I think we could signify that the `name` is optional by making the argument and the corresponding instance variable nillable. That way we could keep the non-empty check for map paths.

spec/marten/routing/map_spec.cr
```diff
@@ -173,6 +173,51 @@ describe Marten::Routing::Match do
...
+      admin_map.path("/articles", article_map)
+
+      map = Marten::Routing::Map.new
+      map.path("/articles", article_map)
+      map.path("/admin", admin_map, name: :admin)
+
+      map.reverse("article:list", {} of String => String).should eq "/articles/list"
+      map.reverse("admin:article:list", {} of String => String).should eq "/admin/articles/list"
+    end
+
```
ellmetha, line 220: I think we may also wany to add a spec covering the fact that the map's namespace gets overridden by the path-level name.

For example:

```crystal
      sub_map = Marten::Routing::Map.new(:inc)
      sub_map.path("/xyz", Marten::Handlers::Base, name: "xyz")

      map = Marten::Routing::Map.new
      map.path("/home", sub_map, :not_inc)

      map.reverse("not_inc:xyz", {} of String => String).should eq "/home/xyz"
```

docs/docs/handlers-and-http/routing.md
```diff
@@ -107,6 +107,29 @@ In the above example, the following URLs would be generated by Marten in additio
 
 As you can see, both the URLs and the route names end up being prefixed respectively with the path and the name specified in the including route.
 
+:::info
+The `name` parameter for included routes is optional, i.e. `path "/articles", Articles::ROUTES` is also valid. Please note that this will increase the possibility of a name collision and it is therefore advisable to prefix the individual paths of the included route,
```
ellmetha, line 111: Can we put the `#path` call example in a dedicated snippet to improve readability?

ellmetha (commented):
Looks good! 🚀 
I left some minor comments, and we will be able to merge the change after we address those!

ellmetha (approved):
LGTM!
Thanks for taking care of this! 😃

PR #115, opened 2023-08-02
Add callbacks to schema handler

### LINKED ISSUE #111: Add callbacks to schema handlers
# Description

Presently, the only way to extend the logic involved when processing valid/invalid schemas as part of handlers (subclasses of `Marten::Handlers::Schema`) is to override the `#process_valid_schema` and `#process_invalid_schema`.

For example:

```crystal
class MyFormHandler < Marten::Handlers::Schema
  schema MySchema
  template_name "app/my_form.html"
  success_route_name "home"

  def process_valid_schema
    # This method is called when the schema is valid.
    # You can decide to do something with the validated data...
    super
  end

  def process_invalid_schema
    # This method is called when the schema is invalid.
    super
  end
end
```

This works fine, but overriding methods from the `Marten::Handlers::Schema` base class is not ideal. Being able to register methods to execute upon processing valid/invalid schemas would be preferable.

In this light, let's introduce the ability to define new kinds of callbacks as part of schema handlers (subclasses of `Marten::Handlers::Schema`):

* `before_schema_validation` - callbacks executed before the schema instance is actually validated
* `after_schema_validation` - callbacks executed after the schema instance validation has been executed
* `after_successful_schema_validation` - callbacks executed after the schema validation succeeded
* `after_failed_schema_validation` - callbacks executed after the schema validation failed

If one of these callbacks returns an HTTP response object (instance of `Marten::HTTP::Response`), then the obtained response should bypass the rest of the schema handler implementation and be returned directly (like this is the case for [standard handler callbacks](https://martenframework.com/docs/handlers-and-http/introduction#callbacks)).

treagod:
This commit adds the callbacks `before_schema_validation`, `after_schema_validation`, `after_successful_schema_validation` and `after_failed_schema_validation`.

Closes #111

### DIFF
--- a/docs/docs/handlers-and-http/generic-handlers.md
+++ b/docs/docs/handlers-and-http/generic-handlers.md
@@ -140,19 +140,19 @@ By default, such a handler will render the configured template when the incoming
 
 If the schema is valid, a temporary redirect is issued by using the URL corresponding to the `#success_route_name` value (although it should be noted that the way to generate this success URL can be overridden by defining a `#success_url` method). By default, the handler does nothing when the processed schema is valid (except redirecting to the success URL).
 
-
 #### Callbacks
 
 Callbacks let you define logics that are triggered before or after the validation of the schema. This allows you to easily intercept validation
-and handle the response independently of the schema validty. Two callbacks are supported: `before_schema_validation`, `after_schema_validation`,
+and handle the response independently of the schema validity. Four callbacks are supported: `before_schema_validation`, `after_schema_validation`,
 `after_successful_schema_validation` and `after_failed_schema_validation`.
 
-All callbacks can optionally return a [`Marten::HTTP::Response`](pathname:///api/dev/Marten/HTTP/Response.html). When a HTTP response is returned
-all pending callbacks and the success/fail response are skipped.
+All callbacks can optionally return a [`Marten::HTTP::Response`](pathname:///api/dev/Marten/HTTP/Response.html). When a HTTP response is returned,
+all following callbacks are skipped and the obtained response is returned directly, thus bypassing responses that might have been
+returned after by the handler.
 
 ##### `before_schema_validation`
 
-`before_schema_validation` callbacks are executed _before_ a schema is checked for validty.
+`before_schema_validation` callbacks are executed _before_ a schema is checked for validity.
 For example, this capability can be leveraged to inspect the incoming request and verify that a user is logged in:
 
 ```crystal
@@ -170,7 +170,7 @@ end
 
 ##### `after_schema_validation`
 
-`after_schema_validation` callbacks are executed right _after_ a schema is checked for validty.
+`after_schema_validation` callbacks are executed right _after_ a schema is checked for validity.
 For example, this capability can be leveraged to add additional cookies to the response:
 
 ```crystal
@@ -188,8 +188,7 @@ end
 
 ##### `after_successful_schema_validation`
 
-`after_successful_schema_validation` callbacks are executed right _after_ a schema is checked for validty and the
-`after_schema_validation` callbacks, but only if the schema wasn't valid.
+`after_successful_schema_validation` callbacks are executed right _after_ a schema is checked for validity (and after possible `after_schema_validation` callbacks), and only if the schema validation was successful.
 For example, this capability can be leveraged to add additional cookies to the response:
 
 ```crystal
@@ -207,8 +206,8 @@ end
 
 ##### `after_failed_schema_validation`
 
-`after_failed_schema_validation` callbacks are executed right _after_ a schema is checked for validty and the
-`after_schema_validation` callbacks, but only if the schema was valid.
+`after_failed_schema_validation` callbacks are executed right _after_ a schema is checked for validity (and after possible
+`after_schema_validation` callbacks), but only if the schema validation failed.
 For example, this capability can be leveraged to add additional cookies to the response:
 
 ```crystal
--- a/spec/marten/handlers/schema_spec.cr
+++ b/spec/marten/handlers/schema_spec.cr
@@ -220,7 +220,7 @@ describe Marten::Handlers::Schema do
     end
   end
 
-  describe "callbacks" do
+  describe "#post" do
     it "executes success callback" do
       request = Marten::HTTP::Request.new(
         ::HTTP::Request.new(
@@ -451,21 +451,6 @@ module Marten::Handlers::SchemaSpec
 
     private def set_foobar
       self.foobar = "set_foobar"
-    end
-  end
-
-  class TestHandlerWithBeforeValidateResponse < Marten::Handlers::Schema
-    before_schema_validation :return_before_schema_validation_response
-    success_url "https://example.com"
-    template_name "specs/handlers/schema/test.html"
-    schema TestSchema
-
-    def get
-      Marten::HTTP::Response.new("Regular response", content_type: "text/plain", status: 200)
-    end
-
-    private def return_before_schema_validation_response
-      Marten::HTTP::Response.new("before_schema_validation response", content_type: "text/plain", status: 200)
     end
   end
 
--- a/spec/marten/handlers/schema_spec/test_schema.cr
+++ b/spec/marten/handlers/schema_spec/test_schema.cr
@@ -3,7 +3,4 @@ module Marten::Handlers::SchemaSpec
     field :foo, :string
     field :bar, :string
   end
-
-  class EmptySchema < Marten::Schema
-  end
 end
--- a/src/marten/handlers/schema.cr
+++ b/src/marten/handlers/schema.cr
@@ -1,5 +1,5 @@
 require "./concerns/record_listing"
-require "./concerns/schema_callbacks"
+require "./schema/callbacks"
 require "./template"
 
 module Marten
@@ -79,10 +79,12 @@ module Marten
 
       def post
         @response = run_before_validation_callbacks
+        return response! if !@response.nil?
 
         valid_schema = schema.valid?
 
         @response ||= run_after_validation_callbacks
+        return response! if !@response.nil?
 
         if valid_schema
           @response ||= run_after_successful_validation_callbacks
--- a/spec/marten/handlers/schema_spec/empty_schema.cr
+++ b/spec/marten/handlers/schema_spec/empty_schema.cr
@@ -1,0 +1,4 @@
+module Marten::Handlers::SchemaSpec
+  class EmptySchema < Marten::Schema
+  end
+end
--- a/src/marten/handlers/schema/callbacks.cr
+++ b/src/marten/handlers/schema/callbacks.cr
@@ -4,7 +4,7 @@ module Marten
     #
     # This module provides the ability to define callbacks that are executed before, after, after success or/and after
     # failed validation of a schema.
-    # They allow to intercept the incoming request and to return early HTTP responses. TwFouro hooks
+    # They allow to intercept the incoming request and to return early HTTP responses. Four hooks
     # are enabled by the use of this module:
     # - `before_schema_validation` callbacks are executed before the validation of the schema
     # - `after_schema_validation` callbacks are executed right after the validation of the schema
@@ -29,11 +29,11 @@ module Marten
         end
       end
 
-      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
+      # Allows to define callbacks that are called before validating the schema.
       #
       # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
       # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and this
-      # custom response will be returned by the handler instead of the success or fail HTTP response.
+      # custom response will be returned by the handler right away.
       macro before_schema_validation(*names)
           {%
             names.reduce(VALIDATION_CALLBACKS[:before_validation]) do |array, name|
@@ -43,11 +43,11 @@ module Marten
           %}
         end
 
-      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
+      # Allows to define callbacks that are called after validating the schema (whether it's valid or not).
       #
       # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
       # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and this
-      # custom response will be returned by the handler instead of the success or fail HTTP response.
+      # custom response will be returned by the handler right away.
       macro after_schema_validation(*names)
           {%
             names.reduce(VALIDATION_CALLBACKS[:after_validation]) do |array, name|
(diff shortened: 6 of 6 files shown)


treagod:
I'm thinking about turning `Schema#process_valid_schema` and `Schema#process_invalid_schema` private. What do you think?

ellmetha:
> I'm thinking about turning `Schema#process_valid_schema` and `Schema#process_invalid_schema` private. What do you think?

I think we should leave those methods public as they are still heavily documented and leveraged in automatically-generated projects (eg. those using the built-in authentication). I think we can avoid introducing breaking changes by keeping them public, plus I don't see any issues with those being public since some advanced users could still want to override them. 🙂

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+    # failed validation of a schema.
+    # They allow to intercept the incoming request and to return early HTTP responses. TwFouro hooks
+    # are enabled by the use of this module:
+    # - `before_schema_validation` callbacks are executed before the validation of the schema
+    # - `after_schema_validation` callbacks are executed right after the validation of the schema
+    # - `after_successful_schema_validation` callbacks are executed right after the validation but
+    #    only if the schema is valid
+    # - `after_failed_schema_validation` callbacks are executed right after the validation but
+    #    only if the schema is not valid
+    module SchemaCallbacks
```
ellmetha, line 15: Can we define this module under a `src/marten/handlers/schema/` folder instead of defining it next to the shared concerns (asking because this concern is really specific to the schema handler)?

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
+module Marten
+  module Handlers
+    # Provides the ability to define schema validation callbacks.
+    #
+    # This module provides the ability to define callbacks that are executed before, after, after success or/and after
+    # failed validation of a schema.
+    # They allow to intercept the incoming request and to return early HTTP responses. TwFouro hooks
```
ellmetha, line 7:
```suggestion
    # They allow to intercept the incoming request and to return early HTTP responses. Four hooks
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+            _finish_validation_callbacks_setup
+          end
+        end
+
+        macro finished
+          _finish_validation_callbacks_setup
+        end
+      end
+
+      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
```
ellmetha, line 32:
```suggestion
      # Allows to define callbacks that are called before validating the schema.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      #
+      # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
+      # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and this
+      # custom response will be returned by the handler instead of the success or fail HTTP response.
```
ellmetha, line 36:
```suggestion
      # custom response will be returned by the handler right away.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      macro before_schema_validation(*names)
+          {%
+            names.reduce(VALIDATION_CALLBACKS[:before_validation]) do |array, name|
+              array << name.id.stringify
+              array
+            end
+          %}
+        end
+
+      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
```
ellmetha, line 46:
```suggestion
      # Allows to define callbacks that are called after validating the schema (whether it's valid or not).
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      #
+      # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
+      # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and this
+      # custom response will be returned by the handler instead of the success or fail HTTP response.
```
ellmetha, line 50:
```suggestion
      # custom response will be returned by the handler right away.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      macro after_schema_validation(*names)
+          {%
+            names.reduce(VALIDATION_CALLBACKS[:after_validation]) do |array, name|
+              array << name.id.stringify
+              array
+            end
+          %}
+        end
+
+      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
```
ellmetha, line 60:
```suggestion
      # Allows to define callbacks that are called after a successful schema validation.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      #
+      # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
+      # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and this
+      # custom response will be returned by the handler instead of the success or fail HTTP response.
```
ellmetha, line 64:
```suggestion
      # custom response will be returned by the handler right away.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      macro after_successful_schema_validation(*names)
+          {%
+            names.reduce(VALIDATION_CALLBACKS[:after_successful_validation]) do |array, name|
+              array << name.id.stringify
+              array
+            end
+          %}
+        end
+
+      # Allows to do define callbacks that are called after executing `Marten::Schema#valid?`.
```
ellmetha, line 74:
```suggestion
      # Allows to define callbacks that are called after a failed schema validation.
```

src/marten/handlers/concerns/schema_callbacks.cr
```diff
@@ -0,0 +1,165 @@
...
+      #
+      # Those callbacks have access to the incoming `#request` object and they can return a custom HTTP response.
+      # If one of such callbacks returns a custom HTTP response, the following callbacks will be skipped and
+      # this custom response will be returned by the handler instead of the success or fail HTTP response.
```
ellmetha, line 78:
```suggestion
      # this custom response will be returned by the handler right away.
```

src/marten/handlers/schema.cr
```diff
@@ -75,7 +78,19 @@ module Marten
       end
 
       def post
-        schema.valid? ? process_valid_schema : process_invalid_schema
+        @response = run_before_validation_callbacks
+
```
ellmetha, line 82: Here I think we could return early and avoid calling `Schema#valid?` if we already have a response. For example:

```suggestion
        @response = run_before_validation_callbacks
        return response! if !@response.nil?

```

src/marten/handlers/schema.cr
```diff
@@ -75,7 +78,19 @@ module Marten
...
+        valid_schema = schema.valid?
+
+        @response ||= run_after_validation_callbacks
```
ellmetha, line 85: Same here! We can avoid the `if valid_schema ... else ...` block completely:

```suggestion
        @response ||= run_after_validation_callbacks
        return response! if !@response.nil?

```

spec/marten/handlers/schema_spec.cr
```diff
@@ -220,6 +220,124 @@ describe Marten::Handlers::Schema do
     end
   end
 
+  describe "callbacks" do
```
ellmetha, line 223:
```suggestion
  describe "#post" do
```

spec/marten/handlers/schema_spec/test_schema.cr
```diff
@@ -3,4 +3,7 @@ module Marten::Handlers::SchemaSpec
     field :foo, :string
     field :bar, :string
   end
+
+  class EmptySchema < Marten::Schema
+  end
```
ellmetha, line 8: Can we define this empty schema class in another dedicated file? 🙏

spec/marten/handlers/schema_spec.cr
```diff
@@ -274,6 +392,143 @@ module Marten::Handlers::SchemaSpec
...
+    private def set_baz
+      self.baz = "set_baz"
+    end
+
+    private def set_foobar
+      self.foobar = "set_foobar"
+    end
+  end
+
+  class TestHandlerWithBeforeValidateResponse < Marten::Handlers::Schema
```
ellmetha, line 457: This class seems to be defined twice!

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
-If the schema is valid, a temporary redirect is issued by using the URL corresponding to the `#success_route_name` value (although it should be noted that the way to generate this success URL can be overridden by defining a `#success_url` method). By default, the handler does nothing when the processed schema is valid (except redirecting to the success URL). That's why it can be helpful to override the `#process_valid_schema` method to implement any logic that should be triggered after a successful schema validation.
+If the schema is valid, a temporary redirect is issued by using the URL corresponding to the `#success_route_name` value (although it should be noted that the way to generate this success URL can be overridden by defining a `#success_url` method). By default, the handler does nothing when the processed schema is valid (except redirecting to the success URL).
+
+
+#### Callbacks
+
+Callbacks let you define logics that are triggered before or after the validation of the schema. This allows you to easily intercept validation
+and handle the response independently of the schema validty. Two callbacks are supported: `before_schema_validation`, `after_schema_validation`,
```
ellmetha, line 147:
```suggestion
and handle the response independently of the schema validity. Four callbacks are supported: `before_schema_validation`, `after_schema_validation`,
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
 
-If the schema is valid, a temporary redirect is issued by using the URL corresponding to the `#success_route_name` value (although it should be noted that the way to generate this success URL can be overridden by defining a `#success_url` method). By default, the handler does nothing when the processed schema is valid (except redirecting to the success URL). That's why it can be helpful to override the `#process_valid_schema` method to implement any logic that should be triggered after a successful schema validation.
+If the schema is valid, a temporary redirect is issued by using the URL corresponding to the `#success_route_name` value (although it should be noted that the way to generate this success URL can be overridden by defining a `#success_url` method). By default, the handler does nothing when the processed schema is valid (except redirecting to the success URL).
+
```
ellmetha, line 142:
```suggestion
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+
+#### Callbacks
+
+Callbacks let you define logics that are triggered before or after the validation of the schema. This allows you to easily intercept validation
+and handle the response independently of the schema validty. Two callbacks are supported: `before_schema_validation`, `after_schema_validation`,
+`after_successful_schema_validation` and `after_failed_schema_validation`.
+
+All callbacks can optionally return a [`Marten::HTTP::Response`](pathname:///api/dev/Marten/HTTP/Response.html). When a HTTP response is returned
```
ellmetha, line 150:
```suggestion
All callbacks can optionally return a [`Marten::HTTP::Response`](pathname:///api/dev/Marten/HTTP/Response.html). When an HTTP response is returned,
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+all pending callbacks and the success/fail response are skipped.
```
ellmetha, line 151:
```suggestion
all following callbacks are skipped and the obtained response is returned directly, thus bypassing responses that might have been returned after by the handler.
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+
+##### `before_schema_validation`
+
+`before_schema_validation` callbacks are executed _before_ a schema is checked for validty.
```
ellmetha, line 155:
```suggestion
`before_schema_validation` callbacks are executed _before_ a schema is checked for validity.
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+
+  private def require_authenticated_user
+    redirect(login_url) unless user_authenticated?(request)
+  end
+end
+```
+
+##### `after_schema_validation`
+
+`after_schema_validation` callbacks are executed right _after_ a schema is checked for validty.
```
ellmetha, line 173:
```suggestion
`after_schema_validation` callbacks are executed right _after_ a schema is checked for validity.
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+  private def add_new_header : Nil
+    response!.headers["X-Foo"] = "Bar"
+  end
+end
+```
+
+##### `after_successful_schema_validation`
+
+`after_successful_schema_validation` callbacks are executed right _after_ a schema is checked for validty and the
+`after_schema_validation` callbacks, but only if the schema wasn't valid.
```
ellmetha, line 192:
```suggestion
`after_successful_schema_validation` callbacks are executed right _after_ a schema is checked for validity (and after possible `after_schema_validation` callbacks), and only if the schema validation was successful.
```

docs/docs/handlers-and-http/generic-handlers.md
```diff
@@ -138,4 +138,88 @@ end
...
+    response!.headers["X-Foo"] = "Bar"
+  end
+end
+```
+
+##### `after_failed_schema_validation`
+
+`after_failed_schema_validation` callbacks are executed right _after_ a schema is checked for validty and the
+`after_schema_validation` callbacks, but only if the schema was valid.
+For example, this capability can be leveraged to add additional cookies to the response:
```
ellmetha, line 212:
```suggestion
`after_failed_schema_validation` callbacks are executed right _after_ a schema is checked for validity (and after possible
`after_schema_validation` callbacks), but only if the schema validation failed.
For example, this capability can be leveraged to add additional cookies to the response:
```

ellmetha (commented):
This looks good! Thanks for taking care of this! 🚀 
I left some comments (mostly typos to fix) that we can look at, and then we should be able to merge this.

docs/docs/handlers-and-http/generic-handlers.md
treagod: Is it really **an response**? I'm pretty sure its **a response**

ellmetha (approved):
Thanks for adding this! 🚀

PR #117, opened 2023-08-03
 Add support for model and schema slug fields

### LINKED ISSUE #48: Add support for model and schema slug fields
# Description

It should be possible to easily define slug fields in models and schemas. Both field implementations should contribute the right validation logic in order to ensure that only valid slug values are accepted.

Slug model field definition would look like this:

```crystal
class TestModel < Marten::Model
  field :identifier, :slug, max_size: 128
end
```

Slug schema field definition would look like this:

```crystal
class TestSchema < Marten::Schema
  field :identifier, :slug
end
```

treagod:
This commit adds slug field for model and schema.

The slug field is an extension of the string field, but has a default maximum size of 50. Also the slug field is indexed by default, as it expected that it will be used to find records.

Closes #48


ellmetha (approved):
LGTM! 🚀 
Thanks for taking care of this!

PR #122, opened 2023-08-12
Added ArchLinux installation instructions

warezgibzzz:
Updated `installation.md` with installation instructions for ArchLinux and derivatives.


ellmetha (approved):
Thanks for providing this!

Issue #120, opened 2023-08-11
Add support for specifying HTTP response status codes using symbols


ellmetha:
# Description

Presently, the only way to specify status codes in the various handler helpers that return HTTP responses (such as `#respond`, `#render`, `#json`, or `#head`) is to specify the raw status code value, expressed as an integer. It would be convenient to be able to specify those response status codes using [Rails-esque symbols](http://www.railsstatuscodes.com/) (with which a lot of developers are used to).

For example, it would be possible to do things like:

```crystal
class MyHandler < Marten::Handler
  def get
    respond "It does not work", status: :bad_request)
  end
end
```

Let's add support for specifying such status code symbols as part of the following handler helper methods:

* `#respond`
* `#render`
* `#json`
* `#head`

ellmetha:
Fixed by #124.

PR #124, opened 2023-09-13
add HTTP Status code lookup class

kates:
Taking a shot at adding this feature https://github.com/martenframework/marten/issues/120 using the mapping from https://crystal-lang.org/api/1.9.2/HTTP/Status.html


spec/marten/http/status_spec.cr
```diff
@@ -0,0 +1,22 @@
...
+describe Marten::HTTP::Status do
+  describe ".status_code" do
+    it "accepts integer status code" do
+      status = Marten::HTTP::Status.status_code(400)
+      status.should eq 400
+    end
+
+    it "accepts symbol status code" do
+      status = Marten::HTTP::Status.status_code(:internal_server_error)
+      status.should eq 500
```
ellmetha, line 12: Would it be possible to iterate over all the <symbol, code> pairs and verify that they produce the expected result?

For example:

```crystal
  [
    [:internal_server_error, 500],
    [:forbidden, 403],
    # ...
  ].each do |code_symbol, expected_status_code|
    Marten::HTTP::Status.status_code(code_symbol).should eq expected_status_code
  end
```

src/marten/handlers/base.cr
```diff
@@ -125,24 +126,36 @@ module Marten
...
       # Returns an empty response associated with a given status code.
-      def head(status : Int32) : HTTP::Response
-        HTTP::Response.new(content: "", content_type: "", status: status)
+      def head(status : StatusCode) : HTTP::Response
+        HTTP::Response.new(
+          content: "",
+          content_type: "",
+          status: HTTP::Status.status_code(status)
+        )
       end
```
ellmetha, line 135: Could we leverage method overloading in order to avoid the extra call to the `HTTP::Status#status_code` method when providing an int?

For example:

```crystal
# Returns an empty response associated with a given status code.
def head(status : Int32) : HTTP::Response
  HTTP::Response.new(content: "", content_type: "", status: status)
end

# :ditto:
def head(status : Symbol) : HTTP::Response
  head(status: HTTP::Status.status_code(status))
end
```

ellmetha (commented):
Thanks a lot for taking care of this! 🚀 

This is definitely a good start! I left a few comments that we can address, and then we should be able to merge the pull request. 🙂

Blacksmoke16:
FWIW this feature isn't really needed. Just type the `status` parameter as `HTTP::Status | Int32 = :ok` and call it a day:

```cr
require "http/status"
 
def respond(status : HTTP::Status | Int32 = :ok)
  pp HTTP::Status.new status
end
 
 
respond HTTP::Status::BAD_REQUEST # => HTTP::Status::BAD_REQUEST
respond 404.                      # => HTTP::Status::NOT_FOUND
respond :im_a_teapot.             # => HTTP::Status::IM_A_TEAPOT
```

ellmetha:
@Blacksmoke16 Thanks a lot for the suggestion! 🙏 This is simpler indeed, definitely the way to go!

@kates could we simplify the PR as suggested by @Blacksmoke16? That way, by using `HTTP::Status` in the type restrictions for the `status` parameter, we wouldn't need to define a `Marten::HTTP::Status` module at all.

ellmetha (approved):
LGTM! Thanks again! 🙇

kates:
Thanks @ellmetha @Blacksmoke16

Issue #110, opened 2023-07-01
Add a template filter to escape string values


ellmetha:
# Description

Let's introduce a new `escape` [template filter](https://martenframework.com/docs/next/templates/introduction#filters) allowing to escape string values in templates. This new filter should probably make use of the [`HTML#escape`](https://crystal-lang.org/api/1.8.2/HTML.html#escape%28string%3AString%2Cio%3AIO%29%3ANil-class-method) method.

For example:

```html
{{ my_var|escape }}
```

If `my_var` was set to `<b>Let's do it</b>`, then the output of the above template would be `&lt;b&gt;Let&#39;s do it&lt;/b&gt;`.

Special care should be taken to ensure that we don't end up applying multiple rounds of escaping to string values that would've been auto-escaped by [Marten's auto-escaping mechanism](https://martenframework.com/docs/next/templates/introduction#auto-escaping). To avoid that, string values escaped with the new `escape` filter should probably be converted to [safe string objects](https://martenframework.com/docs/api/0.3/Marten/Template/SafeString.html).

PR #132, opened 2023-11-22
Add escape filter (#110)

avitkauskas:
Adding escape filter as described in the issue #110.


ellmetha (approved):
Thanks for adding this! 🙏

PR #134, opened 2023-12-15
Add the ability to count records on specific column

### LINKED ISSUE #95: Add the ability to count records using specific columns in query sets
## Description

Presently it is possible to count all the records of a query set by relying on the `#count` method:

```crystal
# Number of published articles:
Article.filter(published: true).count
```

It would be interesting to provide the ability to specify a field name to this method in order to count the number of records for which the specified field has a value in the database:

```crystal
# Number of articles whose subtitle is present:
Article.filter(published: true).count(:subtitle)
```

treagod:
This adds the ability to count records by a specific column name.

Closes #95

### DIFF
--- a/spec/marten/db/query/sql/query_spec.cr
+++ b/spec/marten/db/query/sql/query_spec.cr
@@ -802,7 +802,31 @@ describe Marten::DB::Query::SQL::Query do
         author: student
       )
 
-      Marten::DB::Query::SQL::Query(Marten::DB::Query::SQL::QuerySpec::Article).new.count(:subtitle).should eq 2
+      Marten::DB::Query::SQL::Query(Marten::DB::Query::SQL::QuerySpec::Article).new.count("subtitle").should eq 2
+    end
+
+    it "raises if non existing field is counted" do
+      address = Marten::DB::Query::SQL::QuerySpec::Address.create!(street: "Street 1")
+
+      student = Marten::DB::Query::SQL::QuerySpec::Student.create!(
+        name: "Student 1",
+        email: "student-1@example.com",
+        address: address,
+        grade: "10"
+      )
+
+      Marten::DB::Query::SQL::QuerySpec::Article.create!(
+        title: "Top things 2",
+        subtitle: "Rise of the top things",
+        author: student
+      )
+
+      expect_raises(
+        Marten::DB::Errors::InvalidField,
+        "Unable to resolve 'not_existing' as a field. Valid choices are: author_id, id, title, subtitle."
+      ) do
+        Marten::DB::Query::SQL::Query(Marten::DB::Query::SQL::QuerySpec::Article).new.count("not_existing").should eq 1
+      end
     end
 
     it "returns the expected number of results for a filtered query" do
--- a/spec/marten/db/query/sql/query_spec.cr
+++ b/spec/marten/db/query/sql/query_spec.cr
@@ -780,6 +780,31 @@ describe Marten::DB::Query::SQL::Query do
       Marten::DB::Query::SQL::Query(Tag).new.count.should eq 3
     end
 
+    it "returns the expected number of results for an unfiltered query with specific column defined" do
+      address = Marten::DB::Query::SQL::QuerySpec::Address.create!(street: "Street 1")
+
+      student = Marten::DB::Query::SQL::QuerySpec::Student.create!(
+        name: "Student 1",
+        email: "student-1@example.com",
+        address: address,
+        grade: "10"
+      )
+
+      Marten::DB::Query::SQL::QuerySpec::Article.create!(title: "Top things", author: student)
+      Marten::DB::Query::SQL::QuerySpec::Article.create!(
+        title: "Top things 2",
+        subtitle: "Rise of the top things",
+        author: student
+      )
+      Marten::DB::Query::SQL::QuerySpec::Article.create!(
+        title: "Top things 3",
+        subtitle: "Top things awakening",
+        author: student
+      )
+
+      Marten::DB::Query::SQL::Query(Marten::DB::Query::SQL::QuerySpec::Article).new.count(:subtitle).should eq 2
+    end
+
     it "returns the expected number of results for a filtered query" do
       Tag.create!(name: "ruby", is_active: true)
       Tag.create!(name: "crystal", is_active: true)
--- a/src/marten/db/query/sql/query.cr
+++ b/src/marten/db/query/sql/query.cr
@@ -322,7 +322,6 @@ module Marten
 
               s << "#{Model.db_table}.#{field}"
             end
-
           end
 
           private def build_count_default_subquery_query
--- a/src/marten/db/query/sql/query.cr
+++ b/src/marten/db/query/sql/query.cr
@@ -303,7 +303,7 @@ module Marten
             limit = connection.limit_value(@limit)
 
             sql = build_sql do |s|
-              s << "SELECT COUNT(#{column_name ? column_name.split(".")[-1] : nil})"
+              s << "SELECT COUNT(#{column_name ? column_name.split(".")[-1] : "*"})"
               s << "FROM ("
               s << "SELECT"
 
--- a/src/marten/db/query/sql/query.cr
+++ b/src/marten/db/query/sql/query.cr
@@ -292,23 +292,9 @@ module Marten
 
             sql = build_sql do |s|
               if column_name
-                column_name = column_name.to_s
-                s << "SELECT COUNT(#{column_name})"
+                build_count_column_specific_subquery_query(s, column_name)
               else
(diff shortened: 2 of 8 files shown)


src/marten/db/query/sql/query.cr
```diff
@@ -286,20 +286,15 @@ module Marten
...
-              s << "SELECT"
-
-              if distinct
-                s << connection.distinct_clause_for(distinct_columns)
-                s << columns
+              if column_name
+                build_count_column_specific_subquery_query(s, column_name)
               else
-                s << "#{Model.db_table}.#{Model.pk_field.db_column!}"
+                build_count_default_subquery_query(s)
```
ellmetha, line 297: Would it be possible to avoid fowarding the SQL array to the `#build_count_column_specific_subquery_query` and `#build_count_default_subquery_query` methods and simply append the result of these methods to the SQL array instead?

For example:

```crystal
s <<< if column_name
  build_count_column_specific_subquery_query(scolumn_name)
else
  build_count_default_subquery_query
end
```

src/marten/db/query/sql/query.cr
```diff
@@ -313,6 +308,33 @@ module Marten
...
+
+            s << "SELECT COUNT(#{column_name})"
+            s << "FROM ("
+            s << "SELECT"
+
+            if distinct
+              s << connection.distinct_clause_for(distinct_columns)
+            end
+
+            s << "#{Model.db_table}.#{column_name}"
```
ellmetha, line 322: In order to simplify you could make use of `#build_sql` in this method to replace the use of the `s` argument. That way you would return the subquery string directly:

```crystal
column_name = column_name.to_s
  
build_sql do |s|
  s << "SELECT COUNT(#{column_name})"
  s << "FROM ("
  s << "SELECT"
  
  if distinct
    s << connection.distinct_clause_for(distinct_columns)
  end
  
  s << "#{Model.db_table}.#{column_name}"
end
```

src/marten/db/query/sql/query.cr
```diff
@@ -313,6 +308,33 @@ module Marten
...
+            s << "SELECT COUNT(*)"
+            s << "FROM ("
+            s << "SELECT"
+
+            if distinct
+              s << connection.distinct_clause_for(distinct_columns)
+              s << columns
+            else
+              s << "#{Model.db_table}.#{Model.pk_field.db_column!}"
+            end
```
ellmetha, line 335: The same thing could be done here as well!

src/marten/db/query/sql/query.cr
```diff
@@ -86,8 +86,8 @@ module Marten
             @using.nil? ? Model.connection : Connection.get(@using.not_nil!)
           end
 
-          def count
-            sql, parameters = build_count_query
+          def count(column_name : String | Symbol | Nil = nil)
```
ellmetha, line 89: Can we rename this argument to `field`? 🙏  The reason I am asking this is that this is what will be passed to this method in the end: a field name from which we must determine the corresponding column name.

src/marten/db/query/sql/query.cr
```diff
@@ -86,8 +86,8 @@ module Marten
...
+            sql, parameters = build_count_query(column_name)
```
ellmetha, line 90: We are missing a step here. Indeed, before generating the count SQL query we must first ensure that the "column name" (which is actually a model field name) corresponds to an existing field in the queried model.

In order to do so, we can do something like that:

```crystal
def count(field : String | Nil = nil)
  column_name = if !field.nil?
    field_path = verify_field(raw_field)
    relation_field_path = field_path.select { |field, _r| field.relation? }
    
    if relation_field_path.empty? || field_path.size == 1
      "#{Model.db_table}.#{field_path.first[0].db_column!}"
    else
      join = ensure_join_for_field_path(relation_field_path)
      join.not_nil!.column_name(field_path.last[0].db_column!)
    end
  end

  sql, parameters = build_count_query(column_name)
  
  # ...
```

This will essentially ensure that we:

1. Verify that the field exists (by calling the `#verify_field` method)
2. Use the corresponding DB column name (field names and DB column names do not always match!)
3. Set up a join if the targeted column is part of a relation

Note that `#verify_field` will automatically raise the right exception if the specified field does not exist.

ellmetha (commented):
Thanks for looking into that! 👍 

This is definitely a good start! 🚀 I left a few comments below that we will have to address before merging this PR. The most important thing is that we need to add some logic in order to verify that the targeted field exists. Moreover, we should also update the [`Marten::DB::Query::Set#count`](https://github.com/martenframework/marten/blob/a7e3840488ede4258568cce13e501c89fb4f0ab1/src/marten/db/query/set.cr#L101) so that it's possible to call the `#count` query set method with a field name (updating the `Marten::DB::Query::SQL::Query#count` method only isn't sufficient!).

treagod:
> Thanks for looking into that! 👍
> 
> This is definitely a good start! 🚀 I left a few comments below that we will have to address before merging this PR. The most important thing is that we need to add some logic in order to verify that the targeted field exists. Moreover, we should also update the [`Marten::DB::Query::Set#count`](https://github.com/martenframework/marten/blob/a7e3840488ede4258568cce13e501c89fb4f0ab1/src/marten/db/query/set.cr#L101) so that it's possible to call the `#count` query set method with a field name (updating the `Marten::DB::Query::SQL::Query#count` method only isn't sufficient!).

One more question: There is also a `#count` method for the `raw_set.cr`. Do I need to update these parameters too? Even though the method is not calling `#count` on the sql internally?

ellmetha:
> One more question: There is also a #count method for the raw_set.cr. Should I adapt the API so that the method calls are the same? Even though the method is not calling #count on the sql internally?

@treagod no need! For raw query sets we simply execute the specified raw SQL query and determine the number of records based on the obtained results. We wouldn't be able to do on a count on a specific column in that situation.

src/marten/db/query/sql/query.cr
```diff
@@ -313,6 +320,36 @@ module Marten
             {sql, parameters}
           end
 
+          private def build_count_column_specific_subquery_query(column_name : String)
+            build_sql do |s|
+              # Because a subquery with the name "subquery" is used here, we cut off the table name at the front
+              s << "SELECT COUNT(#{column_name.split(".")[-1]})"
```
ellmetha, line 326: I am realizing that relying on additional methods isn't ideal here because the beginning of the count SQL query is defined in the `#build_count_column_specific_subquery_query` / `#build_count_default_subquery_query` methods while the end is still defined in the `#build_count_query` method. This doesn't make it easy to follow the logic of the generation of this SQL query.

Maybe we could put everything in the `#build_count_query` method? I think this would be simpler to follow. 

For example, something along those lines would work IMHO:

```crystal
private def build_count_query(column_name : String | Nil)
  where, parameters = where_clause_and_parameters
  limit = connection.limit_value(@limit)

  sql = build_sql do |s|
    s << "SELECT COUNT(#{column_name ? column_name.split(".")[-1] : nil})"
    s << "FROM ("
    s << "SELECT"

    if distinct
      s << connection.distinct_clause_for(distinct_columns)
    elsif column_name.nil?
      s << "#{Model.db_table}.#{Model.pk_field.db_column!}"
    end

    if !column_name.nil?
      s << column_name
    end

    s << "FROM #{table_name}"
    s << build_joins
    s << where
    s << "LIMIT #{limit}" unless limit.nil?
    s << "OFFSET #{@offset}" unless @offset.nil?
    s << ") subquery"
  end

  {sql, parameters}
end
```

What do you think?

spec/marten/db/query/sql/query_spec.cr
```diff
@@ -780,6 +780,31 @@ describe Marten::DB::Query::SQL::Query do
...
+      )
+      Marten::DB::Query::SQL::QuerySpec::Article.create!(
+        title: "Top things 3",
+        subtitle: "Top things awakening",
+        author: student
+      )
+
+      Marten::DB::Query::SQL::Query(Marten::DB::Query::SQL::QuerySpec::Article).new.count(:subtitle).should eq 2
+    end
+
```
ellmetha, line 807: Could we also add a spec in order to verify that the `#count` method raises a `Marten::DB::Errors::InvalidField` in case the specified field is unknown? 🙏

ellmetha (commented):
A few questions/comments remaining and we should be able to merge this! 🙂

src/marten/db/query/sql/query.cr
ellmetha: I think the number of ifs is reasonable here, but I am also fine with two completely separate methods that take care of generating two distinct (and complete) SQL queries. What I want to avoid ideally (and in order to improve readability) is having a SQL query whose beginning is generated by a method and whose end is generated by another method.

ellmetha (approved):
Thanks for taking care of this! 🙏

PR #139, opened 2023-12-20
Add #sum to query set

### LINKED ISSUE #96: Add the ability to sum the values of a given column in query sets
# Description

There is presently no easy way to sum the values of a given column - at the database level - by leveraging query sets. In order to palliate this, let's add a new `#sum` method to the base query set implementation. This new method should take the name of a field for which the sum should be computed.

For example:

```crystal
Person.all.sum(:age) # => 1234
```

treagod:
Closes #96


treagod:
Is it possible to retrieve the field type? It would help to return `Int` or `Float` instead of `Int | Float`. In addition it would make it possible to throw an error if a non sumable field is specified

PR #143, opened 2024-01-05
Remove unncessary input field from password update form

treagod:
I accidentally added an unncessary email field to the password update form, which is not required


ellmetha (approved):
Thanks for fixing this! 🙏

PR #149, opened 2024-01-28
Add the ability to compute the average of a given column in query sets

### LINKED ISSUE #144: Add the ability to compute the average of a given column in query sets
# Description

There is presently no easy way to average the values of a given column - at the database level - by leveraging query sets. In order to palliate this, let's add a new `#average` method to the base query set implementation. This new method should take the name of a field for which the average should be computed.

For example:

```crystal
Person.all.average(:age) # => 33.6
```

treagod:
This pull request introduces the ability to calculate the average of a column as shown below

```crystal
# Calculates the average price for all products
Product.average(:price)
# Calculates the average for a subset of products
awesome_products = Product.filter(name__startswith: "Awesome")
awesome_products.average(:price)
```

Closes #144


spec/marten/db/model/querying_spec.cr
```diff
@@ -43,6 +43,16 @@ describe Marten::DB::Model::Querying do
     end
   end
 
+  describe "::average" do
+    it "properly calculates the average" do
+      user = TestUser.create!(username: "jd1", email: "jd1@example.com", first_name: "John", last_name: "Doe")
+      Post.create!(author: user, title: "Example post 1", score: 5.0)
+      Post.create!(author: user, title: "Example post 2", score: 5.0)
+
+      Post.average(:score).should be_close(5.0, 0.00001)
```
ellmetha, line 52: 👍

spec/marten/db/query/set_spec.cr
```diff
@@ -406,6 +406,94 @@ describe Marten::DB::Query::Set do
...
+      )
+
+      expect_raises(
+        Marten::DB::Errors::InvalidField,
+        "Cant calculate the average of 'name' (Marten::DB::Field::String)."
+      ) do
+        Marten::DB::Query::Set(Marten::DB::Query::SetSpec::Product).new.average(:name)
+      end
+    end
+  end
```
ellmetha, line 495: I think we can probably avoid re-defining these tests here since those are testing the `Marten::DB::Query::Set::SQL::Query` abstraction!

src/marten/db/model/querying.cr
```diff
@@ -45,6 +45,11 @@ module Marten
             exists?
           end
 
+          # Returns the average of a field for the current model
```
ellmetha, line 48: Can we put some examples of how to use this method in the doc string too? 🙏

src/marten/db/query/set.cr
```diff
@@ -97,6 +97,11 @@ module Marten
           exists?
         end
 
+        # Returns the average of a field for the current query set.
```
ellmetha, line 100: Same question here!

src/marten/db/query/sql/query.cr
```diff
@@ -74,6 +74,25 @@ module Marten
...
 
+          def average(raw_field : String)
+            field_context = get_field_context(raw_field, Model)
+
+            unless field_context.field.is_a?(Marten::DB::Field::Int) ||
+                   field_context.field.is_a?(Marten::DB::Field::Float)
+              raise Errors::InvalidField.new(
+                "Cant calculate the average of '#{raw_field}' (#{field_context.field.class})."
+              )
+            end
```
ellmetha, line 85: I think here we can avoid hardcoding rules regarding what fields are targeted because technically calculating averages should work for `BigInt` fields too for example. Moreover, some custom fields could also correspond to numeric fields at the database level without inheriting from built-in fields like `Int` or `Float`. Calling `#average` should work for those ideally I think.

In this situation, it's probably safer to avoid this check and let the underlying DB driver return an error if the targeted column cannot be averaged. What do you think?

src/marten/db/query/sql/query.cr
```diff
@@ -74,6 +74,25 @@ module Marten
...
+
+            column_name = solve_field_and_column(raw_field).last
+
+            sql, parameters = build_average_query(column_name)
+            connection.open do |db|
+              result = db.scalar(sql, args: parameters)
+              result ? result.to_s.to_f : 0.0
```
ellmetha, line 92: Shouldn't we return `nil` in case no records are matched by the query set? That way it would be easier to differentiate cases where indeed the average is `0.0` to cases where no records are matched by the query set.

src/marten/db/query/sql/query.cr
```diff
@@ -278,6 +297,29 @@ module Marten
             rows_affected.not_nil!
           end
 
+          private def build_average_query(column_name : String)
+            where, parameters = where_clause_and_parameters
+            limit = connection.limit_value(@limit)
+
+            sql = build_sql do |s|
+              s << "SELECT AVG(#{column_name.split(".")[-1]})"
```
ellmetha, line 305: Why can't we keep the table name or table alias here? If the averaged column appears more than once in a query where multiple tables are joined (for example), this could result in errors returned by the database.

ellmetha (commented):
Thanks for providing this! 🙏 

I left a few comments and questions below. Those are mostly related to small aspects that we can address easily; we should be able to merge this PR soon after they are addressed. 🙂

docs/docs/models-and-databases/reference/query-set.md
```diff
@@ -82,6 +82,20 @@ qset = Article.all # returns a query set matching "all" the records of the Artic
...
+
+Allows calculating the average of a numeric field within the records of a specific model. The `#average` method can be used as a class method from any model class, or it can be used as an instance method from any query set object. When used on a query set, it calculates the average of the specified field for the records in that query set.
+
+For example:
+
+```crystal
+average_price = Product.average(:price) # Calculate the average price of all products
+
+# Calculate the average rating for a specific category of products
+electronic_products = Product.where(category: "Electronics")
```
ellmetha, line 95:
```suggestion
electronic_products = Product.filter(category: "Electronics")
```

src/marten/db/query/sql/query.cr
treagod: Because we are using a subquery and the table name is lost in the outer query. We could of course use a simple query `SELECT AVG(table.column) FROM table`. Would this be sufficient?

ellmetha: I don't think so; the issue is that queries making use of `LIMIT` and `OFFSET` will return no results if we don't use a subquery. Let's keep it that way for now!

spec/marten/db/query/set_spec/models/product.cr
```diff
@@ -0,0 +1,8 @@
+module Marten::DB::Query::SetSpec
+  class Product < Marten::Model
+    field :id, :big_int, primary_key: true, auto: true
+    field :name, :string, max_size: 255
+    field :price, :int
+    field :rating, :float, blank: true, null: true
+  end
+end
```
ellmetha, line 8: I guess we no longer need this model class?

ellmetha (approved):
Thanks for providing this! 🙏

spec/marten/db/query/set_spec/models/product.cr
ellmetha: Removed in d2181c346f239581ce9f0207ab5da70502163d06

PR #150, opened 2024-02-02
Use csrf session

### LINKED ISSUE #66: Add support for session-based CSRF tokens
# Description

Presently, [CSRF tokens](https://martenframework.com/docs/security/csrf) are stored in the browser cookies. This is standard practice, but storing such tokens in session can sometimes be a security requirement. As such, we should find a way to make it possible via a dedicated setting.

## Implementation

- [x] Add a `csrf.use_sessions` boolean setting
- [x] Ensure that the `Marten::Handlers::RequestForgeryProtection` stores the token in the session if the related setting is set

treagod:
This PR adds the option to store the CSRF token in a session. The session key is also changeable, but set to `csrftoken` by default.

Closes #66


ellmetha (commented):
Thanks for adding this! 👍 

This looks good! I left a few minor comments. We should be able to merge this PR after they've been addressed.

ellmetha (approved):

PR #154, opened 2024-02-05
Removes duplicated assignment

treagod:
This line seems to be duplicated with `@distinct = true` on line 167 (169 previously).

If I'm wrong just close this PR


ellmetha (approved):
Thanks for fixing this! 👍

PR #160, opened 2024-02-17
Implement `sum` to query sets

### LINKED ISSUE #96: Add the ability to sum the values of a given column in query sets

treagod:
This PR adds the ability to sum all values for a certain DB column.

Closes #96

### DIFF
--- a/src/marten/db/query/set.cr
+++ b/src/marten/db/query/set.cr
@@ -978,7 +978,8 @@ module Marten
 
         # Returns the sum of a field for the current query set.
         #
-        # Calculates the total sum of values within the specified field for the records included in the query set. For example:
+        # Calculates the total sum of values within the specified field for the records
+        # included in the query set. For example:
         #
         # ```
         # order_items = OrderItem.filter(order_id: 123)
--- a/src/marten/db/query/set.cr
+++ b/src/marten/db/query/set.cr
@@ -976,6 +976,20 @@ module Marten
           raise NotImplementedError.new("#sum is not supported for query sets")
         end
 
+        # Returns the sum of a field for the current query set.
+        #
+        # Calculates the total sum of values within the specified field for the records included in the query set. For example:
+        #
+        # ```
+        # order_items = OrderItem.filter(order_id: 123)
+        # total_price = order_items.sum(:price)
+        # ```
+        #
+        # This would calculate the total cost of all items within order number 123.
+        def sum(field : String | Symbol)
+          @query.sum(field.to_s)
+        end
+
         # :nodoc:
         def to_h
           raise NotImplementedError.new("#to_h is not supported for query sets")
--- a/src/marten/db/query/sql/query.cr
+++ b/src/marten/db/query/sql/query.cr
@@ -206,6 +206,59 @@ module Marten
             !(@limit.nil? && @offset.nil?)
           end
 
+          def sum(raw_field : String)
+            column_name = if !raw_field.nil?
+                            field_path = verify_field(raw_field.to_s)
+                            relation_field_path = field_path.select { |field, _r| field.relation? }
+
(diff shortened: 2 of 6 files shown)


src/marten/db/query/sql/query.cr
```diff
@@ -206,6 +206,59 @@ module Marten
...
+              sum = result.to_s
+
+              return 0 if sum.empty?
+
+              number = sum.to_i?
+              number ? number : sum.to_f
+            end
+          end
+
+          private def build_sum_query(column_name)
```
ellmetha, line 235: Can we move this method closer to the other private methods in this class? 🙏

src/marten/db/query/sql/query.cr
```diff
@@ -206,6 +206,59 @@ module Marten
...
+                            field_path = verify_field(raw_field.to_s)
+                            relation_field_path = field_path.select { |field, _r| field.relation? }
+
+                            if relation_field_path.empty? || field_path.size == 1
+                              "#{Model.db_table}.#{field_path.first[0].db_column!}"
+                            else
+                              join = ensure_join_for_field_path(relation_field_path)
+                              join.not_nil!.column_name(field_path.last[0].db_column!)
+                            end
+                          end
```
ellmetha, line 220: Could we simplify these lines since `raw_field` will always be a `String`?

For example:

```suggestion
            field_path = verify_field(raw_field.to_s)
            relation_field_path = field_path.select { |field, _r| field.relation?
            
            column_name = if relation_field_path.empty? || field_path.size == 1
              "#{Model.db_table}.#{field_path.first[0].db_column!}"
            else
              join = ensure_join_for_field_path(relation_field_path)
              join.not_nil!.column_name(field_path.last[0].db_column!)
            end
```

src/marten/db/query/set.cr
```diff
@@ -976,6 +976,21 @@ module Marten
...
+        #
+        # ```
+        # order_items = OrderItem.filter(order_id: 123)
+        # total_price = order_items.sum(:price)
+        # ```
+        #
+        # This would calculate the total cost of all items within order number 123.
+        def sum(field : String | Symbol)
+          @query.sum(field.to_s)
+        end
```
ellmetha, line 992: Can we also add specs for this method? Especially, verifying that the method can be called with both strings and symbols? 🙏

ellmetha (commented):
Thanks for adding this! This looks good! 
I left a couple comments to improve readability and add some specs too. We can merge after! 🙂

src/marten/db/query/set.cr
ellmetha: Yes! I assume that public methods should always be tested. I always test the logic of methods in the spec associated with the corresponding abstractions, and for methods that simply "delegate" I define a simple spec verifying that we have the right behavior in place.

The comment you are mentioning needs a bit more context though! In the other PR, specs that were testing `Marten::DB::Query::SQL::Query` abstraction were essentially duplicated and redefined at an outer level for the `Marten::DB::Query::Set` abstraction. That's something we want to avoid!

But any public method should be tested anyhow. As you said, this method essentially delegates to the `Marten::DB::Query::SQL::Query` abstraction, so at least one test verifying that the method produces the expected behavior for a simple case (basic sum of a query set) is important in my opinion (like the one you added for the `Querying` concern for example). The other "logic' to test and that is specific to this method is the fact that you can call it with a field name expressed as a symbol or a string and that this produces the same result.

ellmetha (approved):

PR #161, opened 2024-02-17
Add `minimum` and `maximum` methods to query sets

### LINKED ISSUE #146: Add the ability to query the minimum or maximum values of a given column in query sets
# Description

There is presently no easy way to query the minimum/maximum values of a given column - at the database level - by leveraging query sets. In order to palliate this, let's add new `#minimum` and `#maximum` methods to the base query set implementation. These new methods should take the name of a field for which the minimum/maximum value should be queried.

For example:

```crystal
Person.all.minimum(:id) # => 1
Person.all.maximum(:id) # => 145012345
```

These methods should also be available from model classes directly:

```crystal
Person.minimum(:id) # => 1
Person.maximum(:id) # => 145012345
```

treagod:
This PR adds the ability retrieve the minimum/maximum value for a certain DB column.

Closes #146


treagod:
I have noticed that the implementations of aggregate methods are very similar. But I would refactor this in a future commit, to be more precise after min/max and sum are in main

ellmetha (approved):
LGTM! Feel free to merge directly once the spec descriptions have been fixed. 🙂

Issue #167, opened 2024-02-21
Return 422 on validation errors


notramo:
When submitting an invalid form, 200 status is returned with the rendered HTML that contains the error message. However, when using [Turbo](https://turbo.hotwired.dev), the page is not morphed with the new HTML, but the old version is kept. [This issue indicates](https://github.com/hotwired/turbo/issues/246) that it is an intended behavior in Turbo, because form submissions should in most cases return a redirect, 204 (no content), or 4xx error.
From [Turbo docs](https://turbo.hotwired.dev/handbook/drive#redirecting-after-a-form-submission):
> The reason Turbo doesn’t allow regular rendering on 200’s from POST requests is that browsers have built-in behavior for dealing with reloads on POST visits where they present a “Are you sure you want to submit this form again?” dialogue that Turbo can’t replicate. Instead, Turbo will stay on the current URL upon a form submission that tries to render, rather than change it to the form action, since a reload would then issue a GET against that action URL, which may not even exist

Changing error status to 422 would be not only a Turbo workaround, but would also comply with the sandard, since 200 means OK, which indicates that no error happened while the rendered HTML contains error messages.

ellmetha:
This is not a bug so I am going to remove the `Bug` label and use `Discussion` instead. 🙂 

Generic views making use of schemas indeed return a 200 response with the schema in case it is invalid so that the template can re-render the invalid schema and showcase validation errors to end-users. This is because these views simply are built in a way to support classic form display/validation actions initiated by end-users (they re-render the form when it's invalid, and otherwise they redirect to a success route in case it's valid). 

I doubt this default behavior is ever going to change, but we can make it easier to override response codes used in various scenarios by adding a new dedicated class method or something along those lines. In the meantime, one could leverage [schema handler callbacks](https://martenframework.com/docs/handlers-and-http/callbacks#schema-handler-callbacks) to fine-tune the responses returned in case of valid/invalid schemas.

ellmetha:
@notramo Using a specific status code over another is always debatable and subject to interpretation. Some frameworks behave like Marten when processing invalid forms, some others return 422 responses, and some others have different behaviors too. What you are asking for is a breaking change technically. I think I am willing to consider it for the 0.5 release as it seems this would be the most pragmatic approach moving forward (given that it would make it easier to integrate with various tools).

notramo:
What current use cases would it break? Is there some client-side mechanism that requires an OK response in case of error?

notramo:
[...]

PR #168, opened 2024-02-22
Return 422 on invalid form submit

notramo:
Proof of concept for #167.
I tested it locally and everything works. It also fixes [Turbo](https://turbo.hotwired.dev) content update.
The only breaking case is that if some front-end JS code relies on the response status, which is unlikely as it's not an API call, but HTML rendering.


ellmetha (approved):
Thanks for providing this change! 👍 

My main concern is that projects that used the built-in auth generated or that were testing invalid schema handler logic will have their specs failing after they update to the 0.5 release (and this is something that will need to be fixed manually). I'll document this in the release notes.

PR #202, opened 2024-03-06
Fix ameba findings in newly created projects with auth

treagod:


ellmetha (approved):
Thanks for fixing this!

PR #211, opened 2024-03-18
Add `--fingerprint` option to `collectassets`

### LINKED ISSUE #205: Add `--fingerprint` option to `marten collectassets` for cache-busting
# Description

This proposal aims to enhance the marten collectassets command with an optional --fingerprint flag to streamline cache-busting and deployment processes in simple applications.

## Functionality

When the --fingerprint flag is set, the marten collectassets should add a unique fingerprint to the filename of each collected asset.
Also a manifest.json should be generated, which maps the original asset filenames to their fingerprinted counterparts.

Example manifest.json:
```json
{
    "assets/css/app.css": "assets/css/app.3d7ea1.css"
}
```

The generated manifest.json has to be set manually in your Marten configuration (e.g., in `config/settings/production.cr`):

```crystal
Marten.configure :production do |config|
    # Rest of config
    config.assets.manifest = "path/to/manifest.json"
end
```

## Motivation

- **Improved Cache-Busting**: Fingerprinting assets ensures browsers always load the latest versions after deployment and prevents stale cache issues.
- **Simplified Deployment**: This feature can reduce reliance on external asset pipelines for straightforward applications.
- **Optional Behavior**: The `--fingerprint` flag maintains flexibility for developers who prefer their own asset handling.

treagod:
This commit adds a `--fingerprint` option, which automatically adds a fingerprint to the collected assets. In addition a `manifest.json` is created, which maps the original file paths to the fingerprinted paths. 

The `manifest.json` is stored in `src/manifest.json` by default. The new `--manifest-path' option allows you to specify an alternate path where the `manifest.json' should be stored if required.

Closes #205

### DIFF
--- a/docs/docs/assets/introduction.md
+++ b/docs/docs/assets/introduction.md
@@ -93,6 +93,12 @@ Considering the above manifest example, trying to resolve `app/home.css` would p
 Marten.assets.url("app/home.css") # => "/assets/app/home.9495841be78cdf06c45d.css"
 ```
 
+:::info
+Additionally, the collectassets command provides a `--fingerprint` option. Using this option automatically fingerprints the collected assets and generates a `manifest.json` file, which maps the original file paths to their fingerprinted versions.
+
+When the `--fingerprint` option is used, it's important to include the path to the generated "manifest.json" in the appropriate environment.config file, otherwise the collected assets can't be found when the URL is resolved.
+:::
+
 ## Resolving asset URLs
 
 As mentioned previously, assets are collected and persisted in a specific storage. When building HTML [templates](../templates/introduction.md), you will usually need to "resolve" the URL of assets to generate the absolute URLs that should be inserted into stylesheet or script tags (for example).
--- a/docs/docs/development/reference/management-commands.md
+++ b/docs/docs/development/reference/management-commands.md
@@ -36,6 +36,8 @@ Please refer to [Asset handling](../../assets/introduction.md) to learn more abo
 ### Options
 
 * `--no-input` - Does not show prompts to the user
+* `--fingerprint` - Attaches a fingerprint to the collected assets
+* `--manifest-path` - Configures where the manifest.json is stored. Only relevant if `--fingerprint` is activated.
 
 ### Examples
 
--- a/spec/marten/cli/manage/command/collect_assets_spec.cr
+++ b/spec/marten/cli/manage/command/collect_assets_spec.cr
@@ -1,4 +1,6 @@
 require "./spec_helper"
+require "digest/sha256"
+require "json"
 
 describe Marten::CLI::Manage::Command::CollectAssets do
   describe "#run" do
@@ -107,6 +109,41 @@ describe Marten::CLI::Manage::Command::CollectAssets do
       command.handle
 
       File.exists?("spec/assets/css/test.css").should be_true
+    end
+
+    it "copies the assets as expected, adds a fingerprint and creates the correct mapping inside a manifest.json" do
+      stdin = IO::Memory.new("yes")
+      stdout = IO::Memory.new
+
+      original_css_asset_path = "spec/test_project/assets/css/test.css"
+
+      command = Marten::CLI::Manage::Command::CollectAssets.new(
+        options: ["--no-color", "--no-input", "--fingerprint"],
+        stdin: stdin,
+        stdout: stdout
+      )
+
+      sha = Digest::SHA256.new
+      sha.file original_css_asset_path
+      file_digest = sha.hexfinal
+
+      command.handle
+
+      output = stdout.rewind.gets_to_end
+      output.includes?("Copying css/test.#{file_digest}.css...").should be_true
+      output.includes?("Creating manifest.json...").should be_true
+
+      File.exists?("spec/assets/css/test.#{file_digest}.css").should be_true
+      File.exists?("spec/assets/manifest.json").should be_true
+
+      json = File.open("spec/assets/manifest.json") do |file|
+        JSON.parse(file)
+      end
+
+      manifest = json.as_h
+
+      manifest.has_key?("css/test.css").should be_true
+      manifest["css/test.css"].to_s.should eq "css/test.#{file_digest}.css"
     end
   end
 end
--- a/spec/marten/cli/manage/command/collect_assets_spec.cr
+++ b/spec/marten/cli/manage/command/collect_assets_spec.cr
@@ -116,9 +116,10 @@ describe Marten::CLI::Manage::Command::CollectAssets do
       stdout = IO::Memory.new
 
       original_css_asset_path = "spec/test_project/assets/css/test.css"
+      manifest_path = File.expand_path "spec/src/manifest.json"
 
       command = Marten::CLI::Manage::Command::CollectAssets.new(
-        options: ["--no-color", "--no-input", "--fingerprint"],
+        options: ["--no-color", "--no-input", "--fingerprint", "--manifest-path", manifest_path],
         stdin: stdin,
         stdout: stdout
       )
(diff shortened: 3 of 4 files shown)


src/marten/cli/manage/command/collect_assets.cr
```diff
@@ -6,10 +8,21 @@ module Marten
           command_name :collectassets
           help "Collect all the assets and copy them in a unique storage."
 
+          @fingerprint : Bool = false
+          @fingerprint_mapping = Hash(String, String).new
           @no_input : Bool = false
+          @manifest_path : String = File.join Marten.apps.main.class._marten_app_location, "manifest.json"
```
ellmetha, line 14: Is it intended that this file will be created at the root of the project by default (and not in the `src` folder?)?

docs/docs/development/reference/management-commands.md
```diff
@@ -36,6 +36,8 @@ Please refer to [Asset handling](../../assets/introduction.md) to learn more abo
 ### Options
 
 * `--no-input` - Does not show prompts to the user
+* `--fingerprint` - Attaches a fingerprint to the collected assets
+* `--manifest-path` - Configures where the manifest.json is stored. Only relevant if `--fingerprint` is activated.
```
ellmetha, line 40: Can we also mention the default value of this option? 🙏

src/marten/cli/manage/command/collect_assets.cr
```diff
@@ -6,10 +8,21 @@ module Marten
...
 
           def setup
+            on_option("fingerprint", "Add a fingerprint to the collected assets") { @fingerprint = true }
             on_option("no-input", "Do not show prompts to the user") { @no_input = true }
+            on_option_with_arg(
+              "manifest-path",
+              arg: "Filepath",
+              description: "Specify where the manifest file should be stored."
```
ellmetha, line 22: I think we can also add a mention of the default path in the option description! 😊

spec/marten/cli/manage/command/collect_assets_spec.cr
```diff
@@ -108,5 +110,41 @@ describe Marten::CLI::Manage::Command::CollectAssets do
 
       File.exists?("spec/assets/css/test.css").should be_true
     end
+
+    it "copies the assets as expected, adds a fingerprint and creates the correct mapping inside a manifest.json" do
```
ellmetha, line 114: Can we also add a similar unit testing the `--fingerprint` option but without setting the `--manifest-path` option (so that we check that the manifest is created in the expected default folder)?

docs/docs/assets/introduction.md
```diff
@@ -93,6 +93,12 @@ Considering the above manifest example, trying to resolve `app/home.css` would p
 Marten.assets.url("app/home.css") # => "/assets/app/home.9495841be78cdf06c45d.css"
 ```
 
+:::info
+The collectassets command provides a `--fingerprint` option. Using this option automatically fingerprints the collected assets and generates a `manifest.json` file, which maps the original file paths to their fingerprinted versions.
```
ellmetha, line 97: I think we can link to the `collectassets` command reference here! 🙂

docs/docs/assets/introduction.md
```diff
@@ -93,6 +93,12 @@ Considering the above manifest example, trying to resolve `app/home.css` would p
...
+
+When the `--fingerprint` option is used, it's important to include the path to the generated "manifest.json" in the appropriate config file, otherwise the collected assets can't be found when the URL is resolved.
```
ellmetha, line 99: Can we link to the `assets.manifests` setting reference here as well? 🙏

spec/marten/cli/manage/command/collect_assets_spec.cr
```diff
@@ -1,4 +1,6 @@
 require "./spec_helper"
+require "digest/sha256"
+require "json"
```
ellmetha, line 3: We probably don't need to require `json` here as it should already be required by Marten (other JSON operations are already performed by the framework!).

ellmetha (approved):
Thanks for adding this! This looks good! I left some comments and questions in order to improve a few aspects - after those we should be able to merge this! 🙂

ellmetha (commented):
Thanks for adding this! This looks good! I left some comments and questions in order to improve a few aspects - after those we should be able to merge this! 🙂

src/marten/cli/manage/command/collect_assets.cr
treagod: This resolves to `src/manifest.json` for me 🤔 but if we are already talking about it, is there a better variable for it? Because this setting resolves to the `marten/src` instead of the test project directory when running the specs without changing this value.

treagod:
@ellmetha please have a look at my [comment](https://github.com/martenframework/marten/pull/211/files#r1533240248) regarding the app location. After resolving this I'll merge this one!

src/marten/cli/manage/command/collect_assets.cr
```diff
@@ -6,10 +8,26 @@ module Marten
           command_name :collectassets
           help "Collect all the assets and copy them in a unique storage."
 
+          @fingerprint : Bool = false
+          @fingerprint_mapping = Hash(String, String).new
           @no_input : Bool = false
+          @manifest_path : String = File.join Marten.apps.main.class._marten_app_location, "manifest.json"
```
ellmetha, line 14: Can we rewrite this to:

```suggestion
          @manifest_path : String = File.join(Marten.apps.main.class._marten_app_location, "manifest.json")
```

to improve readability?
```

File: src/marten/cli/manage/command/collect_assets.cr [hunk 9]
```diff
@@ -42,16 +77,36 @@ module Marten
...
+
+            if fingerprint? && collected_count > 0
+              FileUtils.mkdir_p(Path[@manifest_path].dirname) # Ensure path exists
+
+              print("  › Creating #{style(@manifest_path, mode: :dim)}...", ending: "")
+              File.open(@manifest_path, "w") do |file|
+                file.print @fingerprint_mapping.to_json
+              end
+              print(style(" DONE", fore: :light_green, mode: :bold))
+            end
```
ellmetha, line 89: Could we move this logic to a new `#create_manifest_file` private method?

src/marten/cli/manage/command/collect_assets.cr
```diff
@@ -42,16 +77,36 @@ module Marten
             if collected_count == 0
               print("No assets to collect...")
             end
+
+            if fingerprint? && collected_count > 0
+              FileUtils.mkdir_p(Path[@manifest_path].dirname) # Ensure path exists
+
+              print("  › Creating #{style(@manifest_path, mode: :dim)}...", ending: "")
```
ellmetha, line 84: Ideally, I think it would be better to show a relative manifest path here in order to ensure. that the command output is not too verbose. What about:

```suggestion
              relative_manifest_path = Path[@manifest_path].relative_to(Marten::Apps::Config.compilation_root_path)
              print("  › Creating #{style(relative_manifest_path, mode: :dim)}...", ending: "")

```

That way we will show the path of the manifest relative to the Marten root folder.

src/marten/cli/manage/command/collect_assets.cr
```diff
@@ -1,3 +1,5 @@
+require "digest/md5"
```
ellmetha, line 1: Marten already requires `digest/md5`, so we don't need to add this requirement here! 😃

spec/marten/cli/manage/command/collect_assets_spec.cr
```diff
@@ -1,4 +1,5 @@
 require "./spec_helper"
+require "digest/md5"
```
ellmetha, line 2: This requirement is not needed technically since Marten already requires `digest/md5`!

ellmetha (commented):
I left a couple of additional comments with some small things to fix, but it looks good! I can do a last review after those are fixed and we should be able to merge this. 🙂

PR #213, opened 2024-03-21
Add queryset macro to `RecordRetrieving` concern

### LINKED ISSUE #180: Add `queryset` macro to handlers retrieving records
# Description

As discussed in #179, let's ensure that the [`#queryset`](https://martenframework.com/docs/api/0.4/Marten/Handlers/RecordListing.html#queryset(queryset)-macro) macro that is available to `Marten::Handlers::RecordList` handlers is also available to handlers that use a query set for retrieving a specific record (like `Marten::Handlers::RecordDetail` handlers for example).

treagod:
Adds a macro to override the queryset for retrieving records in the generic handlers

Closes #180


ellmetha (approved):
LGTM!

PR #216, opened 2024-03-28
Override template loaders

### LINKED ISSUE #15: Allow to explicitly override template loaders
# Description

Right now, Marten templates are configured through the use of the following configuration options:

```crystal
  config.templates.app_dirs = true
  config.templates.cached = true
  config.templates.dirs = [
    Path["src/templates"],
  ]
```

These configuration options results in the initialization of `Marten::Template::Loader::Base` objects that will allow to load the actual templates at runtime. 

It could be interesting to let users define custom template loader classes that can be configured to be used by Marten at runtime in order to load templates. In order to do so, one option would be to define a `config.templates.loaders` configuration that could be used as follows:

```crystal
  config.templates.loaders = [
    Marten::Template::Loader::AppDirs.new,
    Marten::Template::Loader::FileSystem.new("src/templates"),
    # other loaders... 
  ]
```

When used, the `config.templates.loaders` configuration option would have precedence over all the other options (they would be ignored in this situation).

treagod:
Adds the option to manually define the template loaders. When setting `templates.loaders` all other template loader options will be ignored.

Closes #15


docs/docs/development/reference/settings.md
```diff
@@ -763,6 +763,16 @@ config.templates.dirs = [
 ]
 ```
 
+### `loaders`
+
+Default: nil
```
ellmetha, line 768:
```suggestion
Default: `nil`
```

docs/docs/templates/reference/loaders.md
```diff
@@ -0,0 +1,25 @@
...
+description: Template loaders reference.
+---
+
+This page provides a reference for all the available template loaders that can be used to customize template retrieval in Marten.
+
+## FileSystem Loader
+
+**Class**: [`Marten::Template::Loader::FileSystem`](pathname:///api/dev/Marten/Template/Loader/FileSystem.html)
+
+Loads templates directly from the file system.
```
ellmetha, line 12: Can we provide an initialization example here? 😊

docs/docs/templates/reference/loaders.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Class**: [`Marten::Template::Loader::AppDirs`](pathname:///api/dev/Marten/Template/Loader/AppDirs.html)
+
+Coordinates template loading from application directories. Relies on instances of FileSystem.
+
+## Cached Loader
+
+**Class**: [`Marten::Template::Loader::Cached`](pathname:///api/dev/Marten/Template/Loader/Cached.html)
+
+Provides a caching layer for compiled templates. Can wrap other loaders to optimize retrieval.
```
ellmetha, line 24: Same question here: an initialization example would be helpful I think for people looking at this reference doc.

docs/docs/templates/introduction.md
```diff
@@ -238,6 +238,17 @@ Marten.templates.get_template("foo/bar.html")
...
+### Customizing Template Loaders
+
+The [`templates.loaders`](../development/reference/settings.md#loaders) setting provides fine-grained control over how Marten discovers and loads templates. This setting expects an array of template loader classes, which must inherit from `Marten::Template::Loader::Base`. This setting will overwrite the default template loaders
+configured by Marten. This gives you the option to configure [custom loaders](./how-to/create-custom-loaders.md) to load templates from different sources, such as databases or in-memory data structures.
+
+Example: Configuring a File System Loader
+
+```crystal
+config.templates.loaders = [Marten::Template::Loader::FileSystem.new("/path/to/templates")] of Marten::Template::Loader::Base
+```
```
ellmetha, line 250: I think it would be better for the overall structure of this introduction page to put this in an "tip" admonition block. For example:

```suggestion
:::tip Customizing template loaders
The [`templates.loaders`](../development/reference/settings.md#loaders) setting provides fine-grained control over how Marten discovers and loads templates. This setting expects an array of template loader classes, which must inherit from `Marten::Template::Loader::Base`. This setting will overwrite the default template loaders
configured by Marten. This gives you the option to configure [custom loaders](./how-to/create-custom-loaders.md) to load templates from different sources, such as databases or in-memory data structures.

For example, the following can be done to configure a file system loader by making use of this setting:

```crystal
config.templates.loaders = [Marten::Template::Loader::FileSystem.new("/path/to/templates")] of Marten::Template::Loader::Base
```
:::
```

File: docs/docs/templates/how-to/create-custom-loaders.md [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+
+Marten has built-in support for [common template loaders](../reference/loaders.md), but the framework also allows you to write your own template loader that you can leverage as part of your project's templates.
+
+## Defining a template loaders
+
+Template loaders are subclasses of the [`Marten::Template::Loader::Base`](pathname:///api/dev/Marten/Template/Loader/Base.html) abstract class. They must implement a single `#get_template_source` method: this method returns the raw content of a template from a provided template name.
+
+For example, rendering the template `content.html` with a file system loader initialised with `Marten::Template::Loader::FileSystem.new("/app/custom_dir/templates")` would return the content defined in `/app/custom_dir/templates/content.html`.
+
+Let's say we want to write a `DatabaseTemplate` template loader: we first have to define a new class which inherits from `Marten::Template::Loader::Base`. This new class needs to define a `#get_template_source` method which takes a template_name string argument and also returns a string.
```
ellmetha, line 15:
```suggestion
Let's say we want to write a `DatabaseTemplate` template loader: we first have to define a new class which inherits from `Marten::Template::Loader::Base`. This new class needs to define a `#get_template_source` method which takes a `template_name` string argument and also returns a string.
```

docs/docs/templates/how-to/create-custom-loaders.md
```diff
@@ -0,0 +1,29 @@
...
+
+For simplicity we assume that there already exists a model `HtmlTemplate` with a `name` and `content` field:
```
ellmetha, line 17:
```suggestion
For simplicity we assume that there already exists an `HtmlTemplate` model, featuring a `name` field and `content` field:
```

docs/docs/templates/how-to/create-custom-loaders.md
```diff
@@ -0,0 +1,29 @@
...
+
+```crystal
+class DatabaseTemplate < Marten::Template::Loader::Base
+  def get_template_source(template_name) : String
+    begin
+      return HtmlTemplate.get!(name: template_name).content.not_nil!
```
ellmetha, line 23: We can use the `#content!` field method instead of relying on `not_nil!`.

```suggestion
      return HtmlTemplate.get!(name: template_name).content!
```

spec/marten/conf/global_settings_spec.cr
```diff
@@ -245,6 +245,39 @@ describe Marten::Conf::GlobalSettings do
...
+      end
+    end
+
+    it "setup templates uses default loaders when setting is nil" do
+      with_overridden_setting("templates.loaders", nil, nilable: true) do
+        Marten.setup_templates
+        Marten.templates.loaders.size.should eq 2
+      end
+    end
+  end
```
ellmetha, line 279: Can we move these specs to the `spec/marten_spec.cr` file? The reason I am asking this is that the `Marten#setup` or `Marten#setup_templates` methods are part of the top-level `Marten` module, and as such the corresponding specs should live in the associated spec file. 🙂

ellmetha (commented):
This looks good! Thanks for adding the how-to doc related to custom loader creations! I left a couple of comments to improve the documentation and added specs mostly.

PR #218, opened 2024-04-04
Override method

### LINKED ISSUE #215: Implement `MethodOverride` Middleware for HTTP Method Overrides in Forms
# Description

HTML forms natively support only GET and POST requests. To enable RESTful actions like DELETE, PUT, and PATCH, we need a mechanism to translate form submissions into their intended HTTP methods. This would help to resolve #184.

## Proposal

Introduce a `MethodOverride` middleware that intercepts form-data POST requests. This middleware should:

- Check for `_method` Field: Examine if the request contains an input field named `_method`.
- Override Request Method: If `_method` exists, override the request method with the specified value (e.g., DELETE, PUT, PATCH).
- Configurable Allowed Methods: Provide a method like `allowed_override_methods` returning an array of permitted HTTP methods for overriding. This array should be customizable to control which methods can be used.

Example Usage:

```html
<form method="POST">
  <input type="hidden" name="_method" value="DELETE">
</form>
```

## Considerations

Implement tag helper, e.g. `{% form_delete %}`, to help reduce the chance of typos being made.

treagod:
This PR introduces a new Middleware which allows to override the request method if a valid `_method` key is present in the request body.

It also introduces a tag helper which shall prevent typos and make it easier to define the `_method` form input.

Closes #215


treagod:
One more thing: do you see this middleware as a default for Marten? Or should it be up to the project to decide wether the override mechanism is needed?

ellmetha:
> One more thing: do you see this middleware as a default for Marten? Or should it be up to the project to decide wether the override mechanism is needed?

IMHO this should not be one of the default middlewares (in generated projects) since most projects won't need this functionality at all. Each project can decide to use if this becomes necessary! 🙂

ellmetha (commented):
Thanks for starting this! This is definitely a good start! I think we are missing a few things in order to be able to merge this, I'll definitely need to do another pass at reviewing this pull request. 🙂 

Ideally, the way the new middleware behaves should be configurable. Especially (i) the name of the method override input and (ii) the allowed methods that can be overridden. We could add new `method_override.input_name` and `method_override.allowed_methods` settings to make this configurable. I would also add that the `method_override.allowed_methods` should probably default to the `PATCH`, `PUT`, and `DELETE` methods only (same defaults as Rails). What do you think?

Additionally, should we allow the method override to be specified via headers too? I know that Rails supports specifying it in the `X-Http-Method-Override` header. This could be a nice addition for situations where this must be done through headers. Here too, the value of this header should be configurable ideally (eg. `method_override.header_name` setting).

docs/docs/development/reference/settings.md
```diff
@@ -611,6 +611,27 @@ The base URL to use when exposing media files URLs. This base URL will be used b
 This setting is only used if `media_files.storage` is `nil`.
 :::
 
+## Method Overriding Settings
+
+Method overriding settings allow configuring how the [`MethodOverride`](../../handlers-and-http/reference/middlewares#method-override-middleware) Middleware should handle HTTP method overrides in forms. These settings are all available under the `method_override` namespace:
```
ellmetha, line 616:
```suggestion
Method overriding settings allow configuring how the [`MethodOverride`](../../handlers-and-http/reference/middlewares.md#method-override-middleware) Middleware should handle HTTP method overrides in forms. These settings are all available under the `method_override` namespace:
```

spec/marten/template/tag/method_input.cr
```diff
@@ -0,0 +1,39 @@
+require "./spec_helper"
```
ellmetha, line 1: Can we rename this file `method_input_spec.cr`? 🙏

src/marten/conf/global_settings.cr
```diff
@@ -255,6 +255,11 @@ module Marten
         @media_files ||= GlobalSettings::MediaFiles.new
       end
 
+      # Provides access to SSL redirect settings.
```
ellmetha, line 258: Typo here:

```suggestion
      # Provides access to method overriding settings.
      #
      # These setting values will only be used by the `Marten::Middleware::MethodOverride` middleware is enabled.
```

src/marten/conf/global_settings/method_override.cr
```diff
@@ -0,0 +1,33 @@
+module Marten
+  module Conf
+    class GlobalSettings
+      # Allows to configure middleware overriding-related settings.
```
ellmetha, line 4:
```suggestion
      # Allows to configure settings related to the method override middleware.
```

src/marten/http/request.cr
```diff
@@ -121,6 +126,10 @@ module Marten
         @request.method.upcase
       end
 
+      protected def method=(method)
+        @request.method = method
+      end
+
```
ellmetha, line 132: Can we move this method closer to the other `protected` method definitions in this abstraction? 🙏

docs/docs/development/reference/settings.md
```diff
@@ -611,6 +611,27 @@ The base URL to use when exposing media files URLs. This base URL will be used b
 This setting is only used if `media_files.storage` is `nil`.
 :::
 
+## Method Overriding Settings
```
ellmetha, line 614:
```suggestion
## Method overriding settings
```

src/marten/middleware/method_override.cr
```diff
@@ -0,0 +1,42 @@
+module Marten
+  abstract class Middleware
+    # Overrides the HTTP method of a request.
+    #
+    # This middleware overrides the HTTP method of a incoming request that include a `_method` field
+    # (in form data or URL-encoded data) or a `X-Http-Method-Override` header,
```
ellmetha, line 6: The comment mentions that the `X-Http-Method-Override` header is supported but the implementation does not seem to check for this header. Afaics we are only checking for the override param. @treagod we can either add the support for the header later on or as part of this PR, both work fine for me. 🙂

src/marten/middleware/method_override.cr
```diff
@@ -0,0 +1,42 @@
...
+        get_response.call
+      end
+
+      private def allowed?(request)
+        request.post?
+      end
+
+      private def allowed_override_method?(method)
+        Marten.settings.method_override.allowed_methods.includes?(method.upcase)
+      end
```
ellmetha, line 27: Could convert the `allowed_methods` to a local set? This will be more efficient for checking that the method is part of the allowed methods. For example something like:

```suggestion
      private def allowed_override_method?(method)
        allowed_method_sets.includes?(method.upcase)
      end
      
      private def allowed_method_sets
        @allowed_method_sets ||= Marten.settings.method_override.allowed_methods.to_set
      end
```

ellmetha (commented):
We're getting closer! I think mostly some typos need to be fixed and we are missing the support for the `X-Http-Method-Override` header (but we can add this as part of another PR if you prefer - see my comment below). Also left a small improvement suggestion for the middleware itself.

ellmetha (approved):
LGTM! Thanks for adding this!

Issue #224, opened 2024-04-10
Add an `--edge` option to the `new` management command


ellmetha:
# Description

Let's make it possible to generate new projects (or applications) that use the development version of Marten through the use of an `--edge` / `-e` option. When this option is set, the generated project structure should include a `shard.override.yml` file (in addition to the `shard.yml` file) with the following content:

```yaml
dependencies:
  marten:
    github: martenframework/marten
    branch: main
```

This will ensure that the project uses the development version of Marten (`main` branch) when running the `shards install` command ([ref](https://crystal-lang.org/reference/1.12/man/shards/index.html#fixing-dependency-version-conflicts)).

notramo:
I would recommend `--nightly`, or `--dev-preview` as those seem less ambigous.

PR #229, opened 2024-05-23
add edge option to the new command

kates:
* add `shard.override.yml.ecr` to both the `app` and `project` template files
* add `edge` option to the `new` command
* check the `edge` context when including files to be generated

https://github.com/martenframework/marten/issues/224


src/marten/cli/manage/command/new.cr
```diff
@@ -25,6 +26,7 @@ module Marten
               description: "Configure default database (options: mysql/postgresql/sqlite3)") do |db|
               @database = db
             end
+            on_option(:e, :edge, description: "Use latest version from main branch") { @edge = true }
```
ellmetha, line 29:
```suggestion
            on_option(:e, :edge, description: "Use the latest development version of Marten") { @edge = true }
```

src/marten/cli/manage/command/new/templates/app/shard.override.yml.ecr
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  marten:
+    github: martenframework/marten
+    branch: main
```
ellmetha, line 4: Since the content of this file is going to be the same for both app and project repository structures, we could define it only once in the `src/marten/cli/manage/command/new/templates/shared/` folder.

spec/marten/cli/manage/command/new_spec.cr
```diff
@@ -173,6 +173,21 @@ describe Marten::CLI::Manage::Command::New do
       end
     end
 
+    it "creates a new project structure from edge" do
```
ellmetha, line 176:
```suggestion
    it "creates the expected project structure when using the --edge option" do
```

ellmetha, line 176: Can we also add a similar test for the application creation case?

spec/marten/cli/manage/command/new_spec.cr
```diff
@@ -173,6 +173,21 @@ describe Marten::CLI::Manage::Command::New do
...
+      stdout = IO::Memory.new
+
+      command = Marten::CLI::Manage::Command::New.new(
+        options: ["--edge", "project", "dummy_project"],
+        stdout: stdout
+      )
+
+      command.handle
+
+      Marten::CLI::Manage::Command::NewSpec::PROJECT_FILES_WITH_EDGE.each do |path|
```
ellmetha, line 186: I think we can probably avoid defining a new `PROJECT_FILES_WITH_EDGE` constant here (in order to avoid duplications!). Instead, we can probably use the existing `PROJECT_FILES` constant to which we append the additional `shard.override.yml`.

For example:

```suggestion
      (
        Marten::CLI::Manage::Command::NewSpec::PROJECT_FILES +
        ['shard.override.yml']
      ).each do |path|
```

ellmetha (commented):
Thanks for adding this! This looks good! I left a couple suggestions of things to improve below; we should be able to merge the pull request after! 🙂

ellmetha (approved):
Looks good! Thanks for adding this! 🙂

Issue #44, opened 2022-11-11
Add support for raw subqueries


ellmetha:
# Description

The Marten ORM presently allows the create of raw SQL queries for which the whole SQL statement must be provided and whose results are automatically mapped to model instances.

For example:

```crystal
Article.raw("SELECT * FROM articles WHERE title = ?", "hello")
Article.raw("SELECT * FROM articles WHERE title = :title AND updated_at > :updated_at", title: "hello", updated_at: Time.local)
```

For cases where it's not needed to specify model field columns directly, it could be interesting to have the ability to filter records by simply providing a raw subquery predicate (eg. `title = ?`) and the associated parameter(s). When doing so, the resulting query set should allow "chaining" additional filters afterwards if necessary.

For example:

```crystal
Article.all.filter("title = ?", "hello")
Article.all.filter("title = :title AND updated_at > :updated_at", title: "hello", updated_at: Time.local)
```

The advantage of doing so is that the selection of the actual model field columns is left to Marten's ORM, which simplifies the use of raw SQL predicates and avoid having to specify full SQL statements.

PR #230, opened 2024-05-23
Implement raw subqueries

treagod:
This commit adds support for raw SQL queries to the `#filter` method and q expressions.

#44


src/marten/db/query/sql/raw_statement_node.cr
```diff
@@ -0,0 +1,99 @@
...
+            # Escape % characters to be not be considered
+            # during formatting process
+            @statement = @statement.gsub("%", "%%")
+
+            case params
+            when Array(::DB::Any)
+              sanitize_positional_parameters(connection)
+            else
+              sanitize_named_parameters(connection)
+            end
```
ellmetha, line 60: This logic and the underlying methods and constants are duplicated from the `Marten::DB::Query::SQL::RawQuery` abstraction. What do you think about defining this logic into a shared `Marten::DB::Query::SQL::HasSanitizer` concern module?

src/marten/db/query/sql/raw_statement_node.cr
```diff
@@ -0,0 +1,99 @@
...
+      module SQL
+        class RawStatementNode < PredicateNode
+          getter statement
+          getter params
+
+          def initialize(
+            @statement : String,
+            @params = [] of ::DB::Any,
+            @children = [] of PredicateNode,
+            @connector = SQL::PredicateConnector::AND, @negated = false, *args
```
ellmetha, line 13:
```suggestion
            @statement : String,
            @params = [] of ::DB::Any,
            @children = [] of PredicateNode,
            @connector = SQL::PredicateConnector::AND, 
            @negated = false, 
            *args
```

ellmetha, line 13: Will this class be ever instantiated from a statement/params and additional `args`? It doesn't appear to be the case. I guess we could remove this argument altogether. What do you think?

ellmetha (commented):
Thanks for starting this! 🙏 

This looks good and I think we might need to refactor a few things in order to be able to merge this (especially in order to avoid some duplications). I will definitely need to do another review. 🙂

src/marten/db/query/set.cr
```diff
@@ -508,6 +508,89 @@ module Marten
...
+        # query_set.filter("is_published = ?", [true])
+        # ```
+        def filter(query_string : String, params : Array)
+          raw_params = [] of ::DB::Any
+          raw_params += params
+
+          add_query_node(RawNode.new(query_string, raw_params))
+        end
+
+        # # Returns a query set whose records match the given query node object.
```
ellmetha, line 576:
```suggestion
        # Returns a query set whose records match the given subquery and named parameters.
```

src/marten/db/query/set.cr
```diff
@@ -508,6 +508,89 @@ module Marten
...
+        #
+        # ```
+        # query_set = Post.all
+        # query_set.filter("is_published = :published", published: true)
+        # ```
+        def filter(query_string : String, **kwargs)
+          filter(query_string, kwargs.to_h)
+        end
+
+        # Returns a query set whose records match the given query node object.
```
ellmetha, line 558:
```suggestion
        # Returns a query set whose records match the given subquery and positional parameters.
```

src/marten/db/query/set.cr
```diff
@@ -508,6 +508,89 @@ module Marten
...
+        #
+        # ```
+        # query_set = Post.all
+        # query_set.filter("is_published = ?", true)
+        # ```
+        def filter(query_string : String, *args)
+          filter(query_string, args.to_a)
+        end
+
+        # Returns a query set whose records match the given query node object.
```
ellmetha, line 543:
```suggestion
        # Returns a query set whose records match the given subquery and named parameters.
```

src/marten/db/query/set.cr
```diff
@@ -508,6 +508,89 @@ module Marten
...
+        # query_set = Post.all
+        # query_set.filter("is_published = true")
+        # ```
+        def filter(query_string : String)
+          return clone if query_string.empty?
+
+          add_query_node(RawNode.new(query_string))
+        end
+
+        # Returns a query set whose records match the given query node object.
```
ellmetha, line 528:
```suggestion
        # Returns a query set whose records match the given subquery and positional parameters.
```

src/marten/db/query/set.cr
```diff
@@ -508,6 +508,89 @@ module Marten
...
+        # This method enables filtering based on raw SQL conditions, offering greater
+        # flexibility than standard field predicates. It returns a modified `Marten::DB::Query::Set`.
+        #
+        # Example:
+        #
+        # ```
+        # query_set = Post.all
+        # query_set.filter("is_published = true")
+        # ```
+        def filter(query_string : String)
```
ellmetha, line 522: All these new methods should probably be added to the `Marten::DB::Model::Querying` module (so that they can be used from model classes directly). What do you think?

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
...
+      module SQL
+        module Sanitizer
+          private NAMED_PARAMETER_RE        = /(:?):([a-zA-Z]\w*)/
+          private POSITIONAL_PARAMETER_CHAR = '?'
+          private POSITIONAL_PARAMETER_RE   = /#{"\\" + POSITIONAL_PARAMETER_CHAR}/
+
+          private def sanitize_positional_parameters(
+            query : String,
+            params : Array(::DB::Any),
+            connection : Connection::Base? = nil
```
ellmetha, line 13: The `connection` argument should be mandatory here. Otherwise, we can't be sure that the generated parameters will work for the actual database at hand. Is there a rationale for making it optional? From what I can see we have access to a connection object in all places where this module is used.

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
...
+              end
+            end
+
+            {sanitized_query, params}
+          end
+
+          private def sanitize_named_parameters(
+            query : String,
+            params : Hash(String, ::DB::Any),
+            connection : Connection::Base? = nil
```
ellmetha, line 36: Same concern here: the `connection` argument should be mandatory!

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
...
+
+            parameter_offset = 0
+            sanitized_query = query.gsub(POSITIONAL_PARAMETER_RE) do
+              # Conditional placeholder generation
+              if connection
+                parameter_offset += 1
+                connection.parameter_id_for_ordered_argument(parameter_offset)
+              else
+                "%s"
+              end
```
ellmetha, line 27: Here we should always use the `connection` object (if we apply the other comments related to the fact that it should be mandatory):

```suggestion
              parameter_offset += 1
              connection.parameter_id_for_ordered_argument(parameter_offset)
```

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
...
+
+              sanitized_params << params[parameter_match]
+
+              # Conditional placeholder generation
+              if connection
+                parameter_offset += 1
+                connection.parameter_id_for_ordered_argument(parameter_offset)
+              else
+                "%s"
+              end
```
ellmetha, line 58: Here we should always use the `connection` object (if we apply the other comments related to the fact that it should be mandatory):

```suggestion
              parameter_offset += 1
              connection.parameter_id_for_ordered_argument(parameter_offset)
```

src/marten/db/query/sql/raw_query.cr
```diff
@@ -1,10 +1,12 @@
+require "./concerns/sanitizer"
 require "./query"
 
 module Marten
   module DB
     module Query
       module SQL
         class RawQuery(Model)
+          include Sanitizer
```
ellmetha, line 9:
```suggestion
          include Sanitizer

```

src/marten/db/query/sql/raw_statement_node.cr
```diff
@@ -0,0 +1,70 @@
...
+          def to_sql(connection : Connection::Base)
+            # Escape % characters to be not be considered
+            # during formatting process
+            @statement = @statement.gsub("%", "%%")
+
+            case params
+            when Array(::DB::Any)
+              sanitize_positional_parameters(@statement, params.as(Array(::DB::Any)))
+            else
+              sanitize_named_parameters(@statement, params.as(Hash(String, ::DB::Any)))
```
ellmetha, line 63: Related to my other comments, I believe we could forward the `connection` object to the `#sanitize_positional_parameters`  and `#sanitize_named_parameters` methods here!

src/marten/db/query/sql/raw_statement_node.cr
```diff
@@ -0,0 +1,70 @@
+require "./concerns/sanitizer"
+
+module Marten
+  module DB
+    module Query
+      module SQL
+        class RawStatementNode < PredicateNode
```
ellmetha, line 7: Could we rename this class `RawPredicateNode` in order to improve consistency (considering that the parent class is named `PredicatedNode`)? 🙏

src/marten/db/query/sql/raw_statement_node.cr
```diff
@@ -0,0 +1,70 @@
...
+              connector: @connector,
+              negated: @negated,
+              predicates: @predicates.dup
+            )
+          end
+
+          def to_sql(connection : Connection::Base)
+            # Escape % characters to be not be considered
+            # during formatting process
+            @statement = @statement.gsub("%", "%%")
```
ellmetha, line 57: I may be missing something, but why is this needed? Afaik, this is not something that we are doing when preparing raw SQL queries presently.

treagod: Given I have a filter with LIKE this is necessary. Post.all.find("name LIKE 'C%'") won't be correctly formatted ( [`where_clause_and_parameters`](https://github.com/martenframework/marten/blob/main/src/marten/db/query/sql/query.cr#L1117) and will throw a **Malformed format string** error

src/marten/db/query/sql/concerns/sanitizer.cr
treagod: For the raw queries this function will replace the placeholders with a `%s` instead of database specific prepared statement parameter.

The reason for this is, that [`where_clause_and_parameters`](https://github.com/martenframework/marten/blob/main/src/marten/db/query/sql/query.cr#L1117) will format the prepared statement afterwards. For Sqlite and Mysql/Mariadb there won't be a problem, but Postgres uses numbered prepared statement parameters which will potentially yield a wrong order because `Query `does not know about the order of the `RawPredicateNode#to_sql` order.

I'd happy if you have any insights for this ordering problem

treagod:
@ellmetha I'll left two comments regarding the connecting object (which is related to the formatting issue), which is why I didn't touch any of the connection finding

src/marten/db/query/sql/raw_statement_node.cr
ellmetha: Gotcha, that makes sense. Can we add a spec to verify that this behavior is in place? 🙏

src/marten/db/query/sql/concerns/sanitizer.cr
ellmetha: That makes sense. Let's keep it that way, I don't think we could find a way to go around the way the parameters are handled in the main `Marten::DB::Query::SQL::Query` abstraction when building the where clause and parameters.

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
+module Marten
+  module DB
+    module Query
+      module SQL
+        module Sanitizer
+          private NAMED_PARAMETER_RE        = /(:?):([a-zA-Z]\w*)/
+          private POSITIONAL_PARAMETER_CHAR = '?'
+          private POSITIONAL_PARAMETER_RE   = /#{"\\" + POSITIONAL_PARAMETER_CHAR}/
+
+          private def sanitize_positional_parameters(
```
ellmetha, line 10: Can we make this public and define specs for this module specifically? 🙏

src/marten/db/query/sql/concerns/sanitizer.cr
```diff
@@ -0,0 +1,67 @@
...
+                connection.parameter_id_for_ordered_argument(parameter_offset)
+              else
+                "%s"
+              end
+            end
+
+            {sanitized_query, params}
+          end
+
+          private def sanitize_named_parameters(
```
ellmetha, line 33: Same question here: this method should be public and unit-tested ideally. The fact that we created this module also means that it must now be tested. 🙂

src/marten/db/query/set.cr
```diff
@@ -508,7 +508,90 @@ module Marten
...
+        # Returns a query set whose records match the given subquery and named parameters.
+        #
+        # This method enables filtering based on raw SQL conditions, offering greater
+        # flexibility than standard field predicates. It returns a modified `Marten::DB::Query::Set`.
+        #
+        # Example:
+        #
+        # ```
+        # query_set = Post.all
+        # query_set.filter("is_published = ?", {published: true})
```
ellmetha, line 585:
```suggestion
        # query_set.filter("is_published = :published", {published: true})
```

spec/marten/db/query/sql/raw_predicate_node_spec.cr
```diff
@@ -0,0 +1,184 @@
...
+
+      node = Marten::DB::Query::SQL::RawPredicateNode.new(
+        "title = :title",
+        raw_params
+      )
+
+      node.to_sql(Marten::DB::Connection.default).should eq(
+        {"title = %s", ["Example title"]}
+      )
+    end
```
ellmetha, line 182: Do we need this unit test? The predicate initialized here is not negated.

src/marten/db/query/sql/raw_predicate_node.cr
```diff
@@ -0,0 +1,70 @@
...
+
+          getter statement
+          getter params
+
+          def initialize(
+            @statement : String,
+            @params = [] of ::DB::Any,
+            @children = [] of PredicateNode,
+            @connector = SQL::PredicateConnector::AND,
+            @negated = false
```
ellmetha, line 18: I actually wonder if we should not avoid making it possible to initialize raw predicate nodes with `negated`. Because since we don't negate `RawNode` objects, we technically don't need to allow initializing `RawPredicateNode` with `negated: true`. Plus, we also don't use `negated` anywhere for raw predicate nodes. We could simply automatically initialize it to `true` (similarly to what we did for `predicates`). WDYT?

src/marten/db/query/raw_node.cr
```diff
@@ -0,0 +1,40 @@
...
+      class RawNode < Node
+        getter statement
+        getter params
+
+        def initialize(
+          @statement : String,
+          @params = [] of ::DB::Any,
+          @children = [] of Node,
+          @connector = SQL::PredicateConnector::AND,
+          @negated = false
```
ellmetha, line 13: Similarly, I think we can avoid making it possible to initialize raw node objects with `negated: true` since we don't need this capability for raw subqueries.

src/marten/db/query/set.cr
```diff
@@ -508,7 +508,90 @@ module Marten
...
+        # ```
+        def filter(query_string : String, params : Hash | NamedTuple)
+          raw_params = {} of String => ::DB::Any
+          params.each { |k, v| raw_params[k.to_s] = v }
+
+          add_query_node(RawNode.new(query_string, raw_params))
+        end
+
+        # Returns a query set whose records match the given subquery and named parameters.
```
ellmetha, line 594:
```suggestion
        # Returns a query set whose records match the given query node.
```

src/marten/db/query/node.cr
```diff
@@ -78,7 +78,7 @@ module Marten
         end
 
         private def combine(other, conn)
-          combined = self.class.new(connector: conn)
+          combined = Node.new(connector: conn)
```
ellmetha, line 81: Should we use `Node` specifically in the `#-` method as well?

ellmetha:
Thanks for the work on this @treagod! We are getting closer! I left additional comments and small fix suggestions; we should be able to merge this very soon. 🎉

src/marten/db/query/sql/raw_predicate_node.cr
treagod: Yes sounds like the right way to do!

spec/marten/db/query/set_spec.cr
```diff
@@ -1298,6 +1298,245 @@ describe Marten::DB::Query::Set do
...
+
+      qset = Marten::DB::Query::Set(Tag).new.filter("name=:name", name: "crystal")
+
+      qset.to_a.should eq [tag_2]
+    end
+
+    it "raises an error when filtering with a misspelled column in a raw SQL condition", tags: "raw" do
+      Tag.create!(name: "crystal", is_active: true)
+
+      expect_raises(Exception) do
```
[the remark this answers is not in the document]
treagod: We could, but I'm not sure it's useful here. All database types throw their own exception. I didn't see a common ancestor of these database exceptions, so I chose this generic exception type instead of checking each individual database exception.

What do you think?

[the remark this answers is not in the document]
ellmetha: We could define specific blocks for every supported backend and verify that the most precise exception is raised (by wrapping `#expect_raises` calls within `#for_mysql` / `#for_postgresql` / `#for_sqlite` blocks). That way, other types of exceptions wouldn't be "caught" if there is an issue with the underlying code, but that's very unlikely anyway. Let's keep it that way!

ellmetha (approved):
Thanks for adding this!

Issue #221, opened 2024-04-09
Make it possible to configure the max age of assets served by the `AssetServing` middleware


ellmetha:
# Description

Right now the way the `Marten::Middleware::AssetServing` middleware behaves can't be configured. This middleware hardcodes the max age of served assets (3600 seconds), and it is not possible to easily override it if necessary. Let's ensure that this value can be configured easily.

## Proposition

Let's introduce a new `assets.max_age` setting. This setting should allow setting the max age as used in the `Cache-Control` header set by the `Marten::Middleware::AssetServing` middleware, and the default value should be 3600s. It should be made clear in the documentation that this setting only applies when the `Marten::Middleware::AssetServing` middleware is enabled.

PR #231, opened 2024-05-23
Assets max age

kates:
Allow changing the `max-age` value in the settings. It still retain the default `3600s` if `max_age` is not configured.

https://github.com/martenframework/marten/issues/221

### DIFF
--- a/spec/marten/conf/global_settings/assets_spec.cr
+++ b/spec/marten/conf/global_settings/assets_spec.cr
@@ -218,4 +218,19 @@ describe Marten::Conf::GlobalSettings::Assets do
       assets_conf.url.should eq "/assets/url/"
     end
   end
+
+  describe "#max_age" do
+    it "returns the expected default value" do
+      assets_conf = Marten::Conf::GlobalSettings::Assets.new
+      assets_conf.max_age.should eq 3600
+    end
+  end
+
+  describe "#max_age=" do
+    it "allows to set max_age value" do
+      assets_conf = Marten::Conf::GlobalSettings::Assets.new
+      assets_conf.max_age = 7200
+      assets_conf.max_age.should eq 7200
+    end
+  end
 end
--- a/spec/marten/conf/global_settings/assets_spec.cr
+++ b/spec/marten/conf/global_settings/assets_spec.cr
@@ -227,7 +227,7 @@ describe Marten::Conf::GlobalSettings::Assets do
   end
 
   describe "#max_age=" do
-    it "allows to set max_age value" do
+    it "allows to set the max_age value" do
       assets_conf = Marten::Conf::GlobalSettings::Assets.new
       assets_conf.max_age = 7200
       assets_conf.max_age.should eq 7200
--- a/spec/marten/middleware/asset_serving_spec.cr
+++ b/spec/marten/middleware/asset_serving_spec.cr
@@ -299,7 +299,6 @@ describe Marten::Middleware::AssetServing do
           ->{ Marten::HTTP::Response.new("Unknown!", content_type: "text/plain", status: 200) }
         )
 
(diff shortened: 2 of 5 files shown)


spec/marten/middleware/asset_serving_spec.cr
```diff
@@ -280,5 +280,27 @@ describe Marten::Middleware::AssetServing do
...
+              headers: HTTP::Headers{
+                "Host"            => "example.com",
+                "Accept-Encoding" => "gzip, deflate, br",
+              }
+            )
+          ),
+          ->{ Marten::HTTP::Response.new("Unknown!", content_type: "text/plain", status: 200) }
+        )
+
+        response.headers[:"Cache-Control"].should eq "private, max-age=7200"
```
ellmetha, line 302: We should avoid mutating settings in specs. You can make use of the `#with_overridden_setting` method to safely mutate setting values for a specific block instead:

```suggestion
      with_overridden_setting("assets.max_age", 7200) do
        middleware = Marten::Middleware::AssetServing.new
  
        ["BigApp.js", "css/BigApp.css"].each do |path|
          response = middleware.call(
            Marten::HTTP::Request.new(
              ::HTTP::Request.new(
                method: "GET",
                resource: "/assets/#{path}",
                headers: HTTP::Headers{
                  "Host"            => "example.com",
                  "Accept-Encoding" => "gzip, deflate, br",
                }
              )
            ),
            ->{ Marten::HTTP::Response.new("Unknown!", content_type: "text/plain", status: 200) }
          )
  
          response.headers[:"Cache-Control"].should eq "private, max-age=7200"
        end
```

spec/marten/middleware/asset_serving_spec.cr
```diff
@@ -280,5 +280,27 @@ describe Marten::Middleware::AssetServing do
         response.headers[:"Cache-Control"].should eq "private, max-age=3600"
       end
     end
+    it "properly sets the custom Cache-Control header" do
```
ellmetha, line 283:
```suggestion
    it "properly sets the custom Cache-Control header based on the assets.max_age setting value" do
```

src/marten/middleware/asset_serving.cr
```diff
@@ -9,7 +9,7 @@ module Marten
     #
     # It should also be noted that this middleware will automatically compress served assets using GZip or deflate based
     # on the incoming Accept-Encoding header. The middleware also sets the Cache-Control header and specifies a max-age
-    # of 3600s.
+    # of 3600s as default and can be configured in `config.assets.max_age` settings.
```
ellmetha, line 12:
```suggestion
    # of 3600s as default and can be configured in the `config.assets.max_age` setting.
```

spec/marten/conf/global_settings/assets_spec.cr
```diff
@@ -218,4 +218,19 @@ describe Marten::Conf::GlobalSettings::Assets do
...
+
+  describe "#max_age" do
+    it "returns the expected default value" do
+      assets_conf = Marten::Conf::GlobalSettings::Assets.new
+      assets_conf.max_age.should eq 3600
+    end
+  end
+
+  describe "#max_age=" do
+    it "allows to set max_age value" do
```
ellmetha, line 230:
```suggestion
    it "allows to set the max-age value" do
```

ellmetha (approved):
Thanks for adding this! 🙏 Left a couple of minor comments, we should be able to merge this after those are addressed. 😃

spec/marten/middleware/asset_serving_spec.cr
kates: updated

ellmetha:
Thanks for adding this! 🙏

Issue #201, opened 2024-03-05
Add `{{ loop.odd? }}` and `{{ loop.even? }}` properties to the for loop variable


ellmetha:
# Description

In order to make it easier to implement specific templates with for loops, let's ensure that the `loop` contextual variable also provides `odd?` and `even?` attributes.

For example:

```jinja
{% for x in items %}
  {% if loop.even? %}
    Do something...
  {% else %}
    Do something else...
  {% endif %{
{% endfor %}
```

PR #232, opened 2024-05-23
add `odd?` and `even?` to the loop tag

kates:
Used modulo 2 to check for "even-ness". Zero is considered as [even](https://en.wikipedia.org/wiki/Parity_of_zero).

https://github.com/martenframework/marten/issues/201

### DIFF
--- a/spec/marten/template/tag/for/loop_spec.cr
+++ b/spec/marten/template/tag/for/loop_spec.cr
@@ -101,4 +101,32 @@ describe Marten::Template::Tag::For::Loop do
       loop.revindex0.should eq 8
     end
   end
+
+  describe "#odd?" do
+    it "returns true if the index is odd" do
+      loop = Marten::Template::Tag::For::Loop.new(items_size: 10)
+      loop.index = 1
+      loop.odd?.should eq true
+
+      loop.index = 2
+      loop.odd?.should eq false
+
+      loop.index = 3
+      loop.odd?.should eq true
+    end
+  end
+
+  describe "#even?" do
+    it "returns true if the index is even" do
+      loop = Marten::Template::Tag::For::Loop.new(items_size: 10)
+      loop.index = 0
+      loop.even?.should eq true
+
+      loop.index = 1
+      loop.even?.should eq false
+
+      loop.index = 2
+      loop.even?.should eq true
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


src/marten/template/tag/for/loop.cr
```diff
@@ -36,6 +36,14 @@ module Marten
...
             @items_size - @index - 1
           end
+
+          def odd?
+            !even?
+          end
+
+          def even?
+            @index % 2 == 0
+          end
```
ellmetha, line 46: Why not use the existing `#odd?` and `#even?` methods?

```suggestion
          def odd?
            @index.odd?
          end

          def even?
            @index.even?
          end
```

spec/marten/template/tag/for/loop_spec.cr
```diff
@@ -101,4 +101,32 @@ describe Marten::Template::Tag::For::Loop do
...
   end
+
+  describe "#odd?" do
+    it "returns true if the index is odd" do
+      loop = Marten::Template::Tag::For::Loop.new(items_size: 10)
+      loop.index = 1
+      loop.odd?.should eq true
+
+      loop.index = 2
+      loop.odd?.should eq false
```
ellmetha, line 112: I would expect this to be in another spec example (eg. `returns false if the index is not odd`).

spec/marten/template/tag/for/loop_spec.cr
```diff
@@ -101,4 +101,32 @@ describe Marten::Template::Tag::For::Loop do
...
+  end
+
+  describe "#even?" do
+    it "returns true if the index is even" do
+      loop = Marten::Template::Tag::For::Loop.new(items_size: 10)
+      loop.index = 0
+      loop.even?.should eq true
+
+      loop.index = 1
+      loop.even?.should eq false
```
ellmetha, line 126: Same remark here: this could be in its own spec example (`returns false if the index is not even`).

src/marten/template/tag/for/loop.cr
kates: Whoa!! I haven't thought about it. Thanks @ellmetha. I'll make the change.

ellmetha (approved):

ellmetha:
Thanks for adding this!

Issue #145, opened 2024-01-23
Make it possible to configure the content type of responses returned by generic handlers involving templates


ellmetha:
# Description

Generic handlers that make use of the [`Marten::Handlers::Rendering`](https://martenframework.com/docs/api/0.4/Marten/Handlers/Rendering.html) to render templates (like the [`Marten::Handlers::Template`](https://martenframework.com/docs/api/0.4/Marten/Handlers/Template.html) generic handler for example), should make it possible to easily configure the content type associated with the generated response containing the rendered template. By default, this content type should be `text/html` but it should be possible to override this by calling a dedicated class method.

The envisioned API is as follows:

```crystal
class MyHandler < Marten::Handlers::Template
  template_name "app/test.xml"
  content_type "application/xml"
end
```

Special care should be taken to ensure that this class method is added at the [`Marten::Handlers::Rendering`](https://martenframework.com/docs/api/0.4/Marten/Handlers/Rendering.html) concern level in order to ensure that all generic handlers including it also benefit from this capability.

PR #233, opened 2024-05-23
content_type class method to generic handlers

kates:
Added `content_type` class method to the `Rendering` concern which can be used to specify a custom content type for generic handlers.

https://github.com/martenframework/marten/issues/145


ellmetha (approved):
LGTM! Thanks for adding this!

Issue #200, opened 2024-03-05
Add a `csrf_input` template tag


ellmetha:
# Description

Marten already provides a [`csrf_token`](https://martenframework.com/docs/templates/reference/tags#csrf_token) template tag in order to generate the CSRF token value and access it in templates. Let's also introduce a `csrf_input` template tag that will make it easy to generate the following form input containing the CSRF token:

```jinja
<input type="hidden" name="csrftoken" value="{{ csrf_token }}" />
```

This will make it easier to write forms containing a CSRF token input and not bother about possible typos regarding the input name.

PR #234, opened 2024-05-25
csrf-input tag that uses csrf_token as value

kates:
Added a new `csrf_input` tag that would take the output of the `csrf_token` tag and create the html hidden input that uses that output as value.

https://github.com/martenframework/marten/issues/200


ellmetha (approved):
Thanks for adding this! 🙏

PR #249, opened 2024-07-21
Added expires property to the session

Cuspid88:
Now we can set the expiration time for cookies.
```cr
session.expires = {TIME IN SECONDS}
```
Note: _Value 0 is would make the session last as long as the browser is open._

### DIFF
--- a/src/marten/http/session/store/base.cr
+++ b/src/marten/http/session/store/base.cr
@@ -38,6 +38,11 @@ module Marten
             @expires_in ||= Time::Span.new(seconds: Marten.settings.sessions.cookie_max_age)
           end
 
+          def expires_at=(value : Time)
+            @modified = true
+            @expires_in = value - Time.local
+          end
+
           def expires_at
             return nil if expires_at_browser_close
             Time.local + expires_in
--- a/src/marten/http/session/store/base.cr
+++ b/src/marten/http/session/store/base.cr
@@ -9,25 +9,38 @@ module Marten
           @accessed : Bool = false
           @modified : Bool = false
           @session_hash : SessionHash? = nil
-          @expires : Int32? = nil
+          @expires_in : Time::Span? = nil
+          @expires_at_browser_close : Bool = false
 
           getter session_key
+
+          getter expires_at_browser_close
 
           def initialize(@session_key : String?)
           end
 
-          def expires=(value : Int32)
-            @expires = value
+          def expires_at_browser_close=(value : Bool)
             @modified = true
+            @expires_at_browser_close = value
           end
 
-          def expires
-            @expires ||= Marten.settings.sessions.cookie_max_age
+          def expires_in=(value : Time::Span)
+            @modified = true
+            @expires_in = value
+          end
+
+          def expires_in=(value : Int64)
+            @modified = true
+            @expires_in = Time::Span.new(seconds: value)
+          end
+
+          def expires_in
+            @expires_in ||= Time::Span.new(seconds: Marten.settings.sessions.cookie_max_age)
           end
 
           def expires_at
-            return nil if expires == 0
-            Time.local + Time::Span.new(seconds: expires)
+            return nil if expires_at_browser_close
+            Time.local + expires_in
           end
 
           # Returns the value associated with the passed key or raises a `KeyError` exception if not found.
--- a/src/marten/http/session/store/base.cr
+++ b/src/marten/http/session/store/base.cr
@@ -9,10 +9,25 @@ module Marten
           @accessed : Bool = false
           @modified : Bool = false
           @session_hash : SessionHash? = nil
+          @expires : Int32? = nil
 
           getter session_key
 
           def initialize(@session_key : String?)
+          end
+
+          def expires=(value : Int32)
+            @expires = value
+            @modified = true
+          end
+
+          def expires
+            @expires ||= Marten.settings.sessions.cookie_max_age
+          end
+
+          def expires_at
+            return nil if expires == 0
+            Time.local + Time::Span.new(seconds: expires)
           end
 
           # Returns the value associated with the passed key or raises a `KeyError` exception if not found.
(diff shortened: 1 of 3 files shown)


ellmetha:
Hey! Thanks for providing this PR. 

I am not sure to understand how this capability would be useful given that it's already possible to configure the session cookie max age with the existing [`sessions.cookie_max_age`](https://martenframework.com/docs/development/reference/settings#cookie_max_age-1) setting. Could you please give me more details about your use case? Why do you need to set the session cookie max age on a per-session basis?

Additionally, I am not sure that changing the expiry of the session cookie is the best strategy for expiring sessions given that cookies can be tampered with on the client side. A better approach would be to make it possible for session stores to change the expiry of the data _on the server side_. This would involve ensuring that each session store properly makes use of the specified expiry time if any (instead of defaulting to the session cookie max age value).

Cuspid88:
> A better approach would be to make it possible for session stores to change the expiry of the data on the server side.

You are absolutely right - this is exactly how I plan to use the session with Redis. But to control the TTL, the general value of the abstract class will be used.
Here I added the TTL changes Redis store:
  [https://github.com/martenframework/marten-redis-session/pull/1](https://github.com/martenframework/marten-redis-session/pull/1)

Cuspid88:
P.S.
> I am not sure to understand how this capability would be useful given that it's already possible to configure the session cookie max age with the existing sessions.cookie_max_age setting. 

I believe it is extremely important to be able to limit the lifetime of a session (For example, when a user does not check the "remember me" option when logging in). It will also be useful to be able to adjust the TTL on the storage.

ellmetha:
> I believe it is extremely important to be able to limit the lifetime of a session (For example, when a user does not check the "remember me" option when logging in). It will also be useful to be able to adjust the TTL on the storage.

Sessions don't last forever by default. They are always expiring based on the [`sessions.cookie_max_age`](https://martenframework.com/docs/development/reference/settings#cookie_max_age-1) setting value. Obviously, this does not allow for fine-grained session expiry configuration.

src/marten/http/session/store/base.cr
```diff
@@ -9,12 +9,27 @@ module Marten
...
           @modified : Bool = false
           @session_hash : SessionHash? = nil
+          @expires : Int32? = nil
 
           getter session_key
 
           def initialize(@session_key : String?)
           end
 
+          def expires=(value : Int32)
```
ellmetha, line 19: Can we name this method (and the associated instance variable) `#expires_at`?

src/marten/http/session/store/base.cr
```diff
@@ -9,12 +9,27 @@ module Marten
...
+            @expires = value
+            @modified = true
+          end
```
ellmetha, line 22: I would prefer that we persist a `Time` object instead of an integer. Eventually, if we want to make it possible to configure that the session expires in a specific amount of times, then I think we should support an additional `#expires_in` method that would accept a `Time::Span` object as input and that would update the `@expires_at` time instance variable accordingly.

src/marten/http/session/store/base.cr
```diff
@@ -9,12 +9,27 @@ module Marten
...
+
+          def expires
+            @expires ||= Marten.settings.sessions.cookie_max_age
+          end
+
+          def expires_at
+            return nil if expires == 0
```
ellmetha, line 29: We shouldn't allow for non-expiring sessions. Can we remove this from this PR? 🙏 

If users want to set a very big expiry time by default that's fine, but I don't think the framework should allow never-expiring sessions as this is a very bad practice.

src/marten/http/session/store/base.cr
Cuspid88: Sure

src/marten/http/session/store/base.cr
Cuspid88: For cookies, this will mean a short session, and when specifying TTL in the storage, you can specify the value [sessions.cookie_max_age](https://martenframework.com/docs/development/reference/settings#cookie_max_age-1)

For example like this
  https://github.com/martenframework/marten-redis-session/pull/1/commits/8fde4d203a9f2b6a3ef88e124b1b74a89f4a2fdf#diff-6677a4928253a79a8b975ad522b82c587c233b1e0f9d40c9709d9ccd90a44368R68

Thus, this does not imply indefinite storage of the session

src/marten/http/session/store/base.cr
Cuspid88: Ok. Just thought that for a store like Redis, an additional value conversion operation would be required (the TTL is specified as Int there)

src/marten/http/session/store/base.cr
ellmetha: The case of making the session cookie expire when the browser closes should be implemented with a different method in my opinion. How about adding an `#expires_at_browser_close` method that would set that would result in the `#expires_at` method returning `nil`? This would be more transparent for users (vs letting them specify a 0 expiry).

Cuspid88: I agree with you.

Cuspid88:
It seemed to me that storing Time:: Span in the @expires_in field would be more rational. 
What do you think?

ellmetha:
@Cuspid88 Internally I think it's fine to persist a `Time::Span` object.

The only thing is that I think it should be possible to set the expiry using either a `Time` or a `Time::Span` object. For example:

```crystal
session.expires_at = 2.days.from_now
session.expires_in = 2.hours
```

As long as we support this API, we can decide to store the expiry as a `Time::Span` object internally within session stores.

Cuspid88:
Is there anything else I can do?

ellmetha:
> Is there anything else I can do?

@Cuspid88 Can we add unit tests for the changes introduced in this pull request? All the public methods added to the `Marten::HTTP::Session::Store::Base` class should ideally be unit-tested.

ellmetha (approved):
Thanks for adding this @Cuspid88!

PR #250, opened 2024-07-23
Fix context example for template handler

treagod:
All current docs that are showing an example of how to add a context variable via the `before_render` macro were missing a colon before the function specifier


ellmetha (approved):
Thanks for finding and fixing this!

PR #251, opened 2024-07-23
Make record generic handler more unified

treagod:
This update improves the consistency of record generic handlers. 

Key changes:

- Use `Article` as reference record
- Use `Article<action>Schema` naming for schemas
- Always refer to **_articles/_** directory for `template_name`.
- Always make the `model` macro the first configuration in the handler

Considerations I didn't rework: Unify the `success_route_name`.


ellmetha (approved):
LGTM!

PR #254, opened 2024-08-01
Fix text template in email generation

igbanam:


ellmetha (approved):
Thank you for finding and fixing this! 🙏

Issue #155, opened 2024-02-05
Add slugify field option to slug field


treagod:
# Description

Let's enhance the functionality of the slug field in Marten by introducing a new optional slugify option. This option will allow developers to specify another field whose value should be slugified to generate the slug automatically.

## Background

Currently, when defining a slug field in Marten, developers manually set the value of the slug field based on the value of another field, typically by applying a slugification function to that field's value. This process can be repetitive and error-prone, especially when dealing with multiple entities with similar slug generation requirements.

So let's add a slugify option to the slug field configuration, which accepts a string or symbol representing the name of the field whose value should be used to generate the slug. When this option is specified, Marten will automatically generate the slug based on the value of the specified field.

## Functionality

The slugify option will be added to the configuration of the slug field in Marten entity definitions.
It will accept a string or symbol representing the name of the field whose value should be slugified.
Upon saving a record, Marten will automatically generate the slug by applying slugification to the value of the specified field.
If the value of the specified field changes, Marten will update the corresponding slug accordingly.

Example Usage:

```crystal
class Article < Marten::Model
  field :id, :big_int, primary_key: true, auto: true
  field :title, String
  field :slug, :slug, slugify: :title
end
```

In this example, the value of the title field will be used to generate the slug for the slug field automatically.

## Considerations

A way to let the developer define his own slugify function

ellmetha:
That would be a neat functionality! 👍 

A few things that come to mind and that we should probably consider:

1. Did you have something in mind regarding how to implement this? I guess generating a `before_validation` callback when the option is used could work.
2. In terms of slug generation, I think we should try to avoid adding a dependency just for that and implement a `Slugifiable` concern (or similar) in the framework directly if possible.

treagod:
I think I was a little too quick in writing the issue. I'm no longer quite sure to what extent we can generate the slug automatically. I'll give it some thought, but I'd leave the issue here with the "Discussion" label for now so that we don't lose sight of it. A standard way to generate the slug would be neat in any case

treagod:
So my solution was to create a concern Sluggable:

```crystal
module Sluggable
  macro included
    macro slugify(from_field, slug_field, *callbacks)
      \{% if callbacks.size == 0 %}
      after_validation :generate_slug
      \{% else %}
      \{% for callback in callbacks %}
      \{{callback.id}} :generate_slug
      \{% end %}
      \{% end %}

      private def generate_slug
          # generate slug logic
      end
    end
  end
end
```

[...]

PR #258, opened 2024-08-08
Add option to automatically generate slug from another field

treagod:
With this change a slug field accepts a `slugify` parameter, which references another field of the same model. When set the slug field we be automatically set to a slugified version of the referenced field.

#155


src/marten/db/field/slug.cr
```diff
@@ -14,11 +20,27 @@ module Marten
           @null = false,
           @unique = false,
           @index = true,
-          @db_column = nil
+          @db_column = nil,
+          @slugify : Symbol? = nil,
+          @slugify_cb : (::String -> ::String) = ->(value : ::String) { generate_slug(value) }
```
ellmetha, line 25: I am uncertain about the need for this field option. If a user wants to customize how slugs are generated, it might be more appropriate for them to write a `before_validation` callback and implement their own slug generation logic. In this situation, what about removing this 
`slugify_cb` option and enforcing the use of the framework's slugify capabilities instead?

src/marten/db/field/slug.cr
```diff
@@ -29,8 +51,18 @@ module Marten
...
-          # No-op max_size automatic checks...
+        private def generate_slug(value)
+          suffix = "-#{Random::Secure.hex(4)}"
+
+          slug = value.gsub(NON_ALPHANUMERIC_RE, "").downcase
+          slug = slug.gsub(WHITESPACE_HYPHEN_RE, "-").strip("-_")
+          slug = slug.gsub(/[^\x00-\x7F]/, "")
+
+          slug[...(max_size - suffix.size)] + suffix
+        end
```
ellmetha, line 62: It would be interesting to extract any slug generation logic in a dedicated concern module (eg. `Marten::Core::Sluggable`). This would allow it to be easily reused in other parts of the framework but also by users of the framework. WDYT?

src/marten/db/field/slug.cr
```diff
@@ -14,11 +20,27 @@ module Marten
...
           @index = true,
-          @db_column = nil
+          @db_column = nil,
+          @slugify : Symbol? = nil,
+          @slugify_cb : (::String -> ::String) = ->(value : ::String) { generate_slug(value) }
         )
+          if @slugify
+            @null = true
+            @blank = true
+          end
```
ellmetha, line 30: I don't think that we should set `null: true` if the `slugify` option is null. Doing so will make the field nullable at the database level, but this may not be wanted! A user should be able to use the `slugify` option and have a non-nullable field column!

src/marten/db/field/slug.cr
```diff
@@ -14,11 +20,27 @@ module Marten
...
+
+        macro check_definition(field_id, kwargs)
+          # No-op max_size automatic checks...
         end
 
         def validate(record, value)
+          if slugify?(value)
+            slug = slugify_cb.call(record.get_field_value(slugify.not_nil!).to_s)
+            record.set_field_value(id, slug)
+            return
```
ellmetha, line 41: IMHO we shouldn't return early here because we will skip other validations by doing so.

src/marten/core/sluggable.cr
```diff
@@ -0,0 +1,31 @@
+module Marten
+  module Core
+    # The Sluggable module provides functionality for generating URL-friendly slugs from strings.
+    module Sluggable
+      extend self
+
+      NON_ALPHANUMERIC_RE  = /[^\w\s-]/
+      WHITESPACE_HYPHEN_RE = /[-\s]+/
+      NON_ASCII_RE         = /[^\x00-\x7F]/
```
ellmetha, line 9: Can we make these constants private? 🙏

src/marten/core/sluggable.cr
```diff
@@ -0,0 +1,31 @@
...
+      #
+      # The slug is created by:
+      # 1. Removing non-alphanumeric characters (except whitespace and hyphens).
+      # 2. Converting the string to lowercase.
+      # 3. Replacing sequences of whitespace and hyphens with a single hyphen.
+      # 4. Removing non-ASCII characters.
+      # 5. Truncating the slug to fit within the max_size, minus the size of a randomly generated suffix.
+      # 6. Stripping trailing hyphens and underscores of the slug without suffix, and appending the suffix.
+      def generate_slug(value, max_size)
+        suffix = "-#{Random::Secure.hex(4)}"
```
ellmetha, line 21: Why have a random value appended to the slugified output? I think that the expectation for people using an automatic slugify function would be to have the input value converted to a slug without appending anything else.

For example:

```crystal
slugify("This is a test")
# => "this-is-a-test"
```

IMHO we should leave the responsibility of ensuring that values are unique to developers. Otherwise, we are doing a bit more than just generating a slug.

src/marten/core/sluggable.cr
ellmetha, line 9: What about unicode characters? Ultimately, and specifically for certain locales, generating slug values containing unicode characters makes a lot of sense. Is it something that we should support?

src/marten/db/field/slug.cr
```diff
@@ -3,8 +3,10 @@ require "./string"
 module Marten
   module DB
     module Field
-      # Represents an slug field.
+      # Represents a slug field.
       class Slug < String
+        private getter slugify
```
ellmetha, line 8: Can we move private getter definitions closer to other private methods/constants? 🙏

src/marten/core/sluggable.cr
```diff
@@ -0,0 +1,31 @@
+module Marten
+  module Core
+    # The Sluggable module provides functionality for generating URL-friendly slugs from strings.
+    module Sluggable
+      extend self
```
ellmetha, line 5: Why use `extend self` here? I would expect that abstractions requiring slug generation capabilities would simply include `Marten::Core::Sluggable`.

src/marten/db/field/slug.cr
```diff
@@ -14,11 +16,21 @@ module Marten
...
         end
 
+        macro check_definition(field_id, kwargs)
+          # No-op max_size automatic checks...
+        end
+
         def validate(record, value)
+          if slugify?(value)
+            slug = Core::Sluggable.generate_slug(record.get_field_value(slugify.not_nil!).to_s, max_size)
+            record.set_field_value(id, slug)
```
ellmetha, line 31: This is a side effect: the `#validate` method is responsible for validating field values but here we are mutating the record. Ideally, this shouldn't be the case. 

What about making use of the [`#prepare_save`](https://github.com/martenframework/marten/blob/753d3f2182e07d78eb91020c76e2d827cc8626d3/src/marten/db/field/base.cr#L98-L103) method instead for assigning the generated slug?

src/marten/core/sluggable.cr
treagod: We could do that. But does the router support that? I just made a quick test with a route "/ü" and the router could not find it because it searched for the encoded "/%C3%BC".

I would say this could be done in a seperated issue if someone really needs it?

ellmetha (approved):
Looks good!

Issue #244, opened 2024-07-12
Add a ReferrerPolicy middleware


ellmetha:
# Description

## Context

Right now, Marten does not set the [`Referrer-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) as part of HTTP responses. Let's introduce a new `Marten::Middleware::ReferrerPolicy` middleware to automatically set this header (based on an associated setting value) if it is not already set.

## Implementation

The new `Marten::Middleware::ReferrerPolicy` middleware should only set the `Referrer-Policy` header if it is not already defined (eg. by the handler that produced the response). The value of the `Referrer-Policy` header should be determined based on the value of a new `referrer_policy` top-level setting which should contain the [intended referrer policy directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#directives).

PR #259, opened 2024-08-08
Add middleware for Referrer-Policy

treagod:
This adds a new Referrer-Policy. If active the middleware will add a Referrer-Policy header to the response if not already defined

#244

### DIFF
--- a/src/marten/conf/global_settings.cr
+++ b/src/marten/conf/global_settings.cr
@@ -60,7 +60,7 @@ module Marten
       # Returns a boolean indicating whether multiple processes can bind to the same HTTP server port.
       getter port_reuse
 
-      # Returns the value to use for the Referrer-Policy header.
+      # Returns the default value to use for the Referrer-Policy header.
       #
       # The value of this setting will be used by the `Marten::Middleware::ReferrerPolicy` middleware when inserting the
       # Referrer-Policy header in HTTP responses.
@@ -132,7 +132,7 @@ module Marten
       # Allows to indicate whether multiple processes can bind to the same HTTP server port.
       setter port_reuse
 
-      # Allows to set the value to use for the Referrer-Policy header.
+      # Allows to set the default value to use for the Referrer-Policy header.
       #
       # This value will be used by the `Marten::Middleware::ReferrerPolicy` middleware when inserting the
       # Referrer-Policy header in HTTP responses.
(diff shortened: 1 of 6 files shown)


src/marten/conf/global_settings.cr
```diff
@@ -126,6 +132,12 @@ module Marten
       # Allows to indicate whether multiple processes can bind to the same HTTP server port.
       setter port_reuse
 
+      # Allows to set the value to use for the Referrer-Policy header.
```
ellmetha, line 135:
```suggestion
      # Allows to set the default value to use for the Referrer-Policy header.
```

src/marten/conf/global_settings.cr
```diff
@@ -60,6 +60,12 @@ module Marten
       # Returns a boolean indicating whether multiple processes can bind to the same HTTP server port.
       getter port_reuse
 
+      # Returns the value to use for the Referrer-Policy header.
```
ellmetha, line 63:
```suggestion
      # Returns the default value to use for the Referrer-Policy header.
```

src/marten/conf/global_settings.cr
```diff
@@ -198,6 +210,7 @@ module Marten
         @middleware = Array(Marten::Middleware.class).new
         @port = 8000
         @port_reuse = true
+        @referrer_policy = "strict-origin-when-cross-origin"
```
ellmetha, line 213: I know that `strict-origin-when-cross-origin` is usually the browser default. That said, I am under the impression that the framework should push for a safer value by default (like `same-origin`). WDYT?

src/marten/middleware/referrer_policy.cr
```diff
@@ -0,0 +1,39 @@
...
+    # - origin-when-cross-origin: The full URL is sent as the referrer when performing a same-origin request,
+    #                             but only the origin is sent for cross-origin requests.
+    # - same-origin: The Referer header is sent with same-origin requests, but not with cross-origin requests.
+    # - strict-origin: Only the origin is sent as the referrer, and only for same-origin requests.
+    # - strict-origin-when-cross-origin: The full URL is sent as the referrer when performing a same-origin request,
+    #                                    but only the origin is sent for cross-origin requests.
+    #                                    No referrer information is sent to less secure destinations.
+    # - unsafe-url: The full URL is always sent as the referrer, regardless of the request's security.
+    #
+    # You can configure the desired policy in the `referrer_policy` setting in your application's configuration.
```
ellmetha, line 24:
```suggestion
    # You can configure the desired policy through the use of the `referrer_policy` setting.
```

ellmetha (commented):
Thanks for tackling this! This looks good! I left a few suggestions and one question. We should be able to merge this afterward! 🙏

src/marten/conf/global_settings.cr
treagod: Yes, totally agree!

ellmetha (approved):
LGTM!

Issue #147, opened 2024-01-28
Add a `marten seed` management command


treagod:
# Description

Let's introduce a new management command `marten seed`, to streamline the process of populating a database with initial data for Marten projects.

## Background:

As Marten projects evolve, it becomes increasingly important to have a reliable and efficient method for seeding the database with initial data. This initial data could include sample records, or any other essential information required for the project to function properly.

## Proposal

The `marten seed` command will automate the process of seeding the database by providing a convenient interface for executing predefined seeding tasks.

These tasks could be defined in a `seed.cr` file at the root of each app/project. For example the file could look like below, but 
```crystal
# seed.cr
# Populate the database of this app
module Articles::Seed
  def run
    author = Author.create!(firstname: "John", lastname: "Doe")
    Article.create!(name: "Article 1", author_id: author.id)
    Article.create!(name: "Article 2", author_id: author.id)
  end
end
```

Calling `marten seed` will fill your database with the data from each seed.cr inside your project/apps. 

A `marten seed --app=my-app` could be also beneficial to only run the seed task from one app.

This task would especially useful when reloading the database frequently in development environments or initialize the database in production systems.

ellmetha:
I like the idea! I wonder if we need to allow this mechanism to be declined on a per-application basis. I assume that applications that would need to set up some pre-defined data would resort to data migrations instead of seeds. In my mind, using seeds is more something that project developers might use in order to pre-create records based on all the models provided by the installed applications. In this situation, I am under the impression that a single top-level `seeds.cr` would probably be sufficient?

In terms of format, do we need the `#run` method? I guess we can keep it simple to and assume that the `seed.cr` is a simple script. For example:

```crystal
require "./src/project"

# Use this file to define how your database should be seeded.
# To apply seeds, execute `marten seed`

author = Author.create!(firstname: "John", lastname: "Doe")
Article.create!(name: "Article 1", author_id: author.id)
Article.create!(name: "Article 2", author_id: author.id)
```

treagod:
You're right, a single `seeds.cr` would be sufficient. But my thought was that it could be beneficial to only seed/reset one application. For example I have a **blogging** and a **journey** app. I want to implement a new feature for the **journey** app and therefor only want to reset and re-seed its models. The blogging app does not concern me.

This still wouldn't make the `module AppName::seed` neccessary, but I thought it could be helpful **if** multiple `seed.cr` exists.

But maybe we could go first with the single `seeds.cr` approach and then implement further features

[...]

PR #266, opened 2024-09-06
Add `marten seed` command

treagod:
This commit adds a new `marten seed` command which executes a seed file (seed.cr by default). This can be configured by providing a `-f, --file` argument

#147


src/marten/cli/manage/command/new/templates/app/src/app/seed.cr.ecr
```diff
@@ -0,0 +1,6 @@
+require "./src/project"
```
treagod, line 1: This won't work if a seed.cr, that is not in the root, was defined

src/marten/cli/manage/command/seed.cr
```diff
@@ -0,0 +1,54 @@
...
+
+          private getter seed_path
+
+          private def run_seed_file
+            if File.exists?(seed_path)
+              print "Running seed file at #{seed_path}"
+              begin
+                Process.run("crystal #{seed_path}", shell: true, output: stdout, error: stderr)
+              rescue ex : Exception
+                print "Error running seed file: #{ex.message}"
```
ellmetha, line 35: I don't think we need to handle exceptions here since your `Process#run` call will already forward errors to `stderr`.

src/marten/cli/manage/command/seed.cr
```diff
@@ -0,0 +1,54 @@
...
+              end
+            else
+              print "Seed file not found at #{seed_path}. Generating default seed.cr..."
+              write_seed_file
```
ellmetha, line 39: I am not sure that the `seed` command should generate anything. Ideally, I think that this command should focus on executing the specified seed file (or output an error if it does not exist) while generating seeds should be done by a dedicated generator. WDYT?

treagod: My intention was that the user doesn't have to think about how to (re-)create a seed file if none is found so he doesn't have to look up what file content is required for seeding.

But I'm open to remove this functionality. Alternatively a dedicated flag could create it? What do you think?

ellmetha: The thing is that the user might have made a typo while typing the `seed` command or incorrectly specified it. We can't really know for sure and we certainly don't want to make these kinds of assumptions if they lead to us creating files that pollute their repository or their file system (eg. if they made a type in the seed name, should we really create a corresponding file automatically?). When the specified file does not exist, I think that the best course of action is really to return a clear error that eventually mentions how to generate the seed file (through the use of a dedicated generator).

Additionally, I really want to enforce the use of generators for all our generation-related cases. So it makes sense to redirect users to this command should they need to generate seeds. For this reason, I don't think that the command should accept a generation-specific flag either: on the one hand we have a `seed` command used for running seeds, and on the other we allow generating seeds with generators.

ellmetha (approved):
Thank you for adding this!

PR #269, opened 2024-10-12
Config db from url

### LINKED ISSUE #257: Make it possible to configure databases through connection strings
# Description

Let's make it possible to easily configure databases through connection strings.

## Context

Right now, Marten allows to explicitly configure databases by providing each of the database attributes (such as the database name, host, port user, password, etc). For example:

```crystal
config.database do |db|
  db.backend = :postgresql
  db.host = "localhost"
  db.name = "my_db"
  db.user = "my_user"
  db.password = "my_passport"
end
```

That said, some cloud providers only provide a connection string in order to connect to a specific database (usually in a `DATABASE_URL` environment variable). The only way to connect to such databases with Marten presently involves parsing the connection string and assigning the right database attributes to the right database settings manually.

## Proposition

Let's make this process easier by making it possible to provide a connection string to the `#database` config method. For example:

```crystal
# Default database
config.database ENV.fetch("DATABASE_URL")

# Other database
config.database :other, ENV.fetch("OTHER_DATABASE_URL")
```

It is worth noting that it should still be possible to provide a block to this alternative `#database` method in order to access the `Marten::Conf::GlobalSettings::Database` object and set additional database settings manually:

```crystal
config.database ENV.fetch("DATABASE_URL") do |db|
  db.initial_pool_size = 2
end
```

jpmartinspt:
Add functionality so one can configure a database from a URL.

Fixes #257

### DIFF
--- a/src/marten/conf/global_settings.cr
+++ b/src/marten/conf/global_settings.cr
@@ -256,12 +256,17 @@ module Marten
         @csrf ||= GlobalSettings::CSRF.new
       end
 
-      # Allows to configure a specific database connection for the application.
-      def database(id = DB::Connection::DEFAULT_CONNECTION_NAME, &)
+      def database(id = DB::Connection::DEFAULT_CONNECTION_NAME, url : String | Nil = nil)
+        self.database id, url do |_|
+        end
+      end
+
+      def database(id = DB::Connection::DEFAULT_CONNECTION_NAME, url : String | Nil = nil, &)
         db_config = @databases.find { |d| d.id.to_s == id.to_s }
         not_yet_defined = db_config.nil?
         db_config = Database.new(id.to_s) if db_config.nil?
         db_config.not_nil!.with_target_env(@target_env) do |db_config_with_target_env|
+          db_config.from_url(url) if url
           yield db_config_with_target_env
         end
         @databases << db_config if not_yet_defined
--- a/src/marten/conf/global_settings/database.cr
+++ b/src/marten/conf/global_settings/database.cr
@@ -92,6 +92,52 @@ module Marten
         def initialize(@id : String)
         end
 
+        def from_url(url : String)
+          # URI.parse cant parse 'sqlite://:memory:'
+          if url.starts_with? "sqlite://:memory:"
+            self.backend = "sqlite"
+            self.name = ":memory:"
+            return
+          end
+
+          uri = URI.parse url
+
+          self.backend = uri.scheme
+          self.host = uri.host
+          self.port = uri.port
+          self.user = uri.user
+          self.password = uri.password
+          self.name = uri.scheme == "sqlite" ? uri.host : uri.path[1..]?
+
+          params_map = uri.query_params.to_h
+
+          if !(v = params_map.delete("checkout_timeout").try &.to_f64).nil?
+            self.checkout_timeout = v
+          end
+
+          if !(v = params_map.delete("initial_pool_size").try &.to_i32).nil?
+            self.initial_pool_size = v
+          end
+
+          if !(v = params_map.delete("max_pool_size").try &.to_i32).nil?
+            self.max_pool_size = v
+          end
+
+          if !(v = params_map.delete("max_idle_pool_size").try &.to_i32).nil?
+            self.max_idle_pool_size = v
+          end
+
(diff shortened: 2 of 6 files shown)


src/marten/conf/global_settings.cr
```diff
@@ -256,12 +256,17 @@ module Marten
         @csrf ||= GlobalSettings::CSRF.new
       end
 
-      # Allows to configure a specific database connection for the application.
-      def database(id = DB::Connection::DEFAULT_CONNECTION_NAME, &)
+      def database(id = DB::Connection::DEFAULT_CONNECTION_NAME, url : String | Nil = nil)
```
jpmartinspt, line 259: @ellmetha might be missing something with the overloading as it's my first time fiddling with crystal, but I think we have to stick with `url` as a named parameter here, therefore configuring with `config.database url: "postgres://my_user:my_db@localhost:1234/db"`

src/marten/conf/global_settings/database.cr
```diff
@@ -92,6 +92,35 @@ module Marten
...
+          self.backend = uri.scheme
+          self.host = uri.host
+          self.port = uri.port
+          self.user = uri.user
+          self.password = uri.password
+          self.name = uri.scheme == "sqlite" ? uri.host : uri.path[1..]?
+
+          params_map = uri.query_params.to_h
+          
+          self.checkout_timeout = params_map.delete("checkout_timeout").try &.to_f64 || @checkout_timeout
```
jpmartinspt, line 114: Would you want to raise a specific exception here in case one of these casts fails?

treagod:
Hi! Before reviewing could you kindly fix the failing GitHub QA  check? This should be done after formatting `./src/marten/conf/global_settings/database.cr` 😊

src/marten/conf/global_settings.cr
```diff
@@ -256,12 +256,17 @@ module Marten
         @csrf ||= GlobalSettings::CSRF.new
       end
 
-      # Allows to configure a specific database connection for the application.
```
ellmetha, line 259: Can we keep method descriptions? 🙏

src/marten/conf/global_settings/database.cr
```diff
@@ -92,6 +92,35 @@ module Marten
...
+          end
+
+          uri = URI.parse url
+
+          self.backend = uri.scheme
+          self.host = uri.host
+          self.port = uri.port
+          self.user = uri.user
+          self.password = uri.password
+          self.name = uri.scheme == "sqlite" ? uri.host : uri.path[1..]?
```
ellmetha, line 110: It's worth noting that `sqlite3` is a valid scheme for SQLite databases. We should definitely account for that!

src/marten/conf/global_settings/database.cr
ellmetha, line 114: It would probably be safer to assign these values only when they are present in the query params:

```suggestion
          if params_map.has_key?("checkout_timeout")
            self.checkout_timeout = params_map.delete("checkout_timeout").try &.to_f64
          end
```

src/marten/conf/global_settings.cr
jpmartinspt: Absolutely, accidental removal :)

src/marten/conf/global_settings/database.cr
jpmartinspt: Unless I'm missing something, the compiler does not like this assignment as it still thinks delete could return a `Nil`.

jpmartinspt: I could also just deal with the deletion of these keys when setting `self.options` if that sounds better.

ellmetha: > Unless I'm missing something, the compiler does not like this assignment as it still thinks delete could return a Nil.

Yes, you have to check that you are in the presence a non-nil value. For example:

```suggestion
          if !(v = params_map.delete("checkout_timeout").try &.to_f64).nil?
            self.checkout_timeout = v
          end
```

jpmartinspt:
@treagod @ellmetha I believe I've addressed all the comments. Feel free to give the review another pass.

ellmetha (approved):
Thank you for adding this! 🙏

Issue #265, opened 2024-09-05
Avoid Unnecessary Database Query on Back Relations


treagod:
# Description

When using back relations in Marten, unnecessary database queries are being executed when accessing the parent relation in a loop, Marten queries the database for each item, even though the parent object (e.g., list) has already been fetched. This results in inefficiencies and extra database load.

**Example Scenario:**

```crystal
# many-to-one relation
List.get!(id: 1).items.each do |item|
  puts item.list # This invokes a database query for each item to fetch the list
end
```

In this scenario, `item.list` should reference the already-fetched list from `List.get!(id: 1)` without triggering additional queries.

## Benefits

- **Improved Performance**: Reduces the number of redundant database queries, improving the performance of applications using Marten.
- **Resource Efficiency**: Lowers the load on the database.
- **Developer Experience**: Aligns with expected behavior, where related objects should efficiently reference each other without extra queries.

PR #275, opened 2024-10-30
Avoid Unnecessary Database Query on Back Relations

treagod:
With this update, child objects will now reference the already-fetched parent, avoiding redundant queries and improving performance.

Closes https://github.com/martenframework/marten/issues/265

### DIFF
--- a/src/marten/db/query/related_set.cr
+++ b/src/marten/db/query/related_set.cr
@@ -5,7 +5,12 @@ module Marten
     module Query
       # Represents a query set resulting from a many-to-one relation.
       class RelatedSet(M) < Set(M)
-        def initialize(@instance : Marten::DB::Model, @related_field_id : String, query : SQL::Query(M)? = nil)
+        def initialize(
+          @instance : Marten::DB::Model,
+          @related_field_id : String,
+          query : SQL::Query(M)? = nil,
+          @assign_related : Bool = false
+        )
           @query = if query.nil?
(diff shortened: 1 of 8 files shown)


src/marten/db/query/related_set.cr
```diff
@@ -15,6 +15,11 @@ module Marten
                    end
         end
 
+        protected def fetch
+          super
+          @result_cache.not_nil!.each { |r| r.assign_related_object(@instance, @related_field_id) }
```
ellmetha, line 20: I don't think that this should be the default behaviour for this query set abstraction. The issue I see here is that we are making the assumption that the `Marten::DB::Query::RelatedSet` abstraction is (or will always be) used for reverse relations only and that setting this object automatically will always make sense, but that may not necessarily be the case! A library could well make use of this abstraction for creating query sets in different contexts, and preloading associated objects might not necessarily be wanted in these situations.

My recommendation would be to:

1. Add a new `assign_related` bool argument to the `#initialize` method of this abstraction that would default to `false`.
2. Make sure that `assign_related: true` is specified when creating the query sets corresponding to reverse many-to-one relations [here](https://github.com/martenframework/marten/blob/c28cd38c38f9ff941c803768051af6a31c21914e/src/marten/db/field/many_to_one.cr#L205).
3. Make sure that the related object instance is assigned only when `assign_related` is set to `true`.

ellmetha (approved):

Issue #273, opened 2024-10-24
Add a localize template tag


ellmetha:
# Description

Let's introduce a new `localize` template tag allowing to perform basic localization operations in templates.

## Context

Marten relies on [crystal-i18n](https://crystal-i18n.github.io/) in order to power its internationalization features. The template engine presently provides a [`translate`](https://martenframework.com/docs/templates/reference/tags#translate) tag to perform translations within templates. That said, there is no easy way to perform [localizations](https://crystal-i18n.github.io/localization.html) in templates at the moment (that is, without having to implement a custom template tag).

## Proposition

Let's introduce a new `localize` tag template that will allow to perform the localizations that are [already provided by crystal-i18n](https://crystal-i18n.github.io/localization.html) (that is, localizing dates, datetimes, and numbers).

Concretely, it should be possible to localize such values using the default format:

```jinja
{% localize 100000 %}
{% localize created_at %}
```

We should also make it possible for users to explicitly define the localization format that they want to use with a dedicated `format` argument:

```jinja
{% localize 100000 format: "custom" %}
{% localize created_at format: "short" %}
```

PR #277, opened 2024-11-19
Add `localize` template tag

treagod:
This PR adds a `localize` template tag which allows the user to localize Numbers, Dates and DateTimes

#273

### DIFF
--- a/docs/docs/templates/reference/tags.md
+++ b/docs/docs/templates/reference/tags.md
@@ -284,7 +284,7 @@ Templates that are included using the `include` template are parsed and rendered
 ## `localize`
 
 The `localize` template tag allows to perform translation lookups by using the  of the project. It must take at least one argument (the translation key) followed by keyword arguments.
-The localize template tag allows performing localization on values such as dates, numbers, and times by using the [I18n gem](https://crystal-i18n.github.io/localization.html) of the project. It must take at least one argument (the value to localize) followed by an optional format keyword argument.
+The `localize` template tag allows performing localization of values such as dates, numbers, and times by using the [I18n gem](https://crystal-i18n.github.io/localization.html), which is leveraged by Marten for its [internationalization features](../../i18n/introduction.md). It must take at least one argument (the value to localize) followed by an optional `format` keyword argument.
 
 For example, the following lines are valid usages of the localize tag:
 
--- a/docs/docs/templates/reference/tags.md
+++ b/docs/docs/templates/reference/tags.md
@@ -292,7 +292,7 @@ For example, the following lines are valid usages of the `localize` tag:
 {% localize price format: "currency" %}
 ```
 
-Values and the format parameter can be resolved as template variables too, but they can also be defined as literal values if necessary. The format argument must match a key defined in the locale file.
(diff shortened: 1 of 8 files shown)


src/marten/template/tag/localize.cr
```diff
@@ -0,0 +1,99 @@
...
+          @kwargs = {} of String => FilterExpression
+          extract_kwargs(kwargs_parts.join(' ')).each do |key, value|
+            @kwargs[key] = FilterExpression.new(value)
+          end
+        end
+
+        def render(context : Context) : String
+          value = @value_expression.resolve(context).raw
+
+          if format_value = @kwargs.delete("format")
```
ellmetha, line 54: Why are we deleting the `format` from the `kwargs` instance variable here? This could make successive uses of the template tag behave inconsistently!

If believe we want to perform a regular key lookup instead:

```suggestion
          if !(format_value = @kwargs["format")]?).nil?
```

src/marten/template/tag/localize.cr
```diff
@@ -0,0 +1,99 @@
...
+
+        private def localize_value(value, format : String?) : String
+          case value
+          when Time, Float64, Int32, Int64
+            valid_value = value.as(Float64 | Int32 | Int64 | Time)
+            format.nil? ? I18n.l(valid_value) : I18n.l(valid_value, format)
+          when Array(Marten::Template::Value)
+            if value.size != 3
+              raise Errors::UnsupportedValue.new(
+                "Localization requires an Array with exactly 3 elements, but received #{value.size} elements. " +
```
ellmetha, line 76:
```suggestion
                "Localization of dates requires an Array with exactly 3 elements, but received #{value.size} elements. " +
```

src/marten/template/tag/localize.cr
```diff
@@ -0,0 +1,99 @@
...
+      # within templates. It can take one mandatory argument (the value to localize) and an optional `format`
+      # keyword argument to specify a localization format.
+      #
+      # Usage examples:
+      # ```
+      # {% localize created_at %}
+      # {% localize 100000 format: "short" %}
+      # ```
+      #
+      # Optionally, the result of the localization can be assigned to a variable using `as` keyword:
```
ellmetha, line 18: Can we specify that the specified format must be one of the formats defined in the locale files too? 🙏

docs/docs/templates/reference/tags.md
```diff
@@ -281,6 +281,30 @@ Additionally, it is important to note that the accessibility of outer context va
...
+The localize template tag allows performing localization on values such as dates, numbers, and times by using the [I18n gem](https://crystal-i18n.github.io/localization.html) of the project. It must take at least one argument (the value to localize) followed by an optional format keyword argument.
+
+For example, the following lines are valid usages of the localize tag:
+
+```html
+{% localize created_at %}
+{% localize price format: "currency" %}
+```
+
+Values and the format parameter can be resolved as template variables too, but they can also be defined as literal values if necessary. The format argument must match a key defined in the locale file.
```
ellmetha, line 296:
```suggestion
The provided values and the `format` argument can be resolved as template variables too, but they can also be defined as literal values if necessary. The `format` argument must match a key defined in the locale file.
```

ellmetha (approved):
Looks good!

Issue #13, opened 2021-07-07
Add support for custom through models for many-to-many fields


ellmetha:
# Description

Right now an intermediary join table is automatically created for each many-to-many field definition. Ideally we should try to support the use of many-to-many fields involving custom through models, which can be necessary if other fields need to be defined as part of the many-to-many relationship.

The issue with a custom through model is that it would be necessary to introspect the specified model in order to identify what is the "origin" foreign key and the "destination" foreign key. Ideally we should find a way to ensure that this is something that is explicitly defined as part of the many-to-many field definition:

```crystal
class Post < Marten::Model
  # Other fields...

  field :tags, :many_to_many, to: Tag, through: CustomPostTags, source: :field1, target: :field2
end
```

miry:
@ellmetha I found it is very usefull and prepared a small PR: https://github.com/martenframework/marten/pull/282

PR #282, opened 2025-01-11
db: add support for custom through models in many-to-many fields

miry:
This update allows specifying a custom class for the `through` option in many-to-many relationships, providing greater flexibility in defining intermediate models.

Example:

```crystal
class Post < Marten::Model
  # Other fields...

  field :tags, :many_to_many, to: Tag, through: CustomPostTags
end
```

This makes it easier to define and manage custom behavior or attributes within the intermediary model.

Reference: https://github.com/martenframework/marten/issues/13

### DIFF
--- a/src/marten/db/field/many_to_many.cr
+++ b/src/marten/db/field/many_to_many.cr
@@ -96,7 +96,9 @@ module Marten
 
             {% field_id_string = field_id.stringify %}
 
-            {% through_model_name = "#{model_klass}#{field_id_string.capitalize.id}" %}
+            {% through_model_name = kwargs[:through] && kwargs[:through].stringify %}
+            {% through_model_name ||= "#{model_klass}#{field_id_string.capitalize.id}" %}
+
             {% through_related_name = "#{from_model_name.downcase.id}_#{field_id_string.downcase.id}" %}
             {% through_from_related_name = through_related_name %}
             {% through_to_related_name = through_related_name %}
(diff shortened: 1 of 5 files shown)


src/marten/db/field/many_to_many.cr
```diff
@@ -120,8 +122,8 @@ module Marten
               register_field(
                 {{ @type }}.new(
                   {{ field_id.stringify }},
-                  {% unless kwargs.is_a?(NilLiteral) %}**{{ kwargs }}{% end %},
```
miry, line 123: This line is not related to the main topic, but I think comma should be present inside the  block. It worked because kwargs is not nil, because `to` is required.

ellmetha:
Hey! 👋

Thanks for attempting to fix this! Unfortunately, I don't think that this solution could fit into the framework in its current state. The main difficulty with implementing this (which I was referring to in #13) is that we need to make sure we can define custom through models while also ensuring that every other piece of functionality related to many-to-many fields keeps working as expected. This includes:

1. The ability to join through many-to-many fields that define a custom through model.
2. The ability to prefetch many-to-many relations that involve a custom through model.

With the proposed changes, it would be possible to indeed have custom "through" model defined, but trying to access the associated records would not necessarily work since the actual "origin" and "destination" foreign key names (in the through model) would not necessarily match the expected ones ([ref](https://github.com/martenframework/marten/blob/57f422b498141ee6be2938b1ee3aa757ac0264d5/src/marten/db/field/many_to_many.cr#L105-L113)). 

We can't rely only on conventions to address this: the framework must be able to handle through models with relations that don't necessarily match the default relation names.

Additionally, there is also the question of creating new records through many-to-many query sets. Right now, it is very easy to add new records to a m2m relation with the `#add` method:

```crystal
tag_1 = Tag.create!(label: "Tag 1")
tag_2 = Tag.create!(label: "Tag 2")
tag_3 = Tag.create!(label: "Tag 3")

article = Article.create!(title: "My article")
article.tags.add(tag_1)
```

What if the custom through model involves mandatory fields? We will need to make it possible to specify custom fields to assign to the through model, otherwise, these capabilities become essentially useless.

miry:
I am closing for now. I am going to work on some thing from the list.

Issue #264, opened 2024-09-02
Make it possible to specify a custom query set when prefetching records


ellmetha:
# Description

Let's make it possible to configure a custom query set to use when prefetching records.

## Context

In some situations, it may be necessary to ensure that records are prefetched using a specific query set. This would be the case when:

* A custom filtering must be applied and a prefetching operation is required.
* A custom ordering  must be applied and a prefetching operation is required.

## Proposition

To allow for that, let's introduce an alternative version of the [`#prefetch`](https://martenframework.com/docs/models-and-databases/reference/query-set#prefetch) method that takes a single relation name as input and an optional query set object:

```crystal
List.prefetch(:items, query_set: Item.order(:position))
```

The configured prefetch operation should take into account the specified query set if one is specified; otherwise, the default one should be used. Special care should be taken to ensure that the type of the query set type is consistent with the specified relation.

PR #288, opened 2025-01-24
Add custom query for prefetching

treagod:
This change introduces the option to provide a custom query for `#prefetch`, e.g.:

```crystal
qs = List.prefetch(:items, Item.order(:position))
```

The type of custom query must match the related type, otherwise an exception is thrown.

#264

### DIFF
--- a/docs/docs/models-and-databases/reference/query-set.md
+++ b/docs/docs/models-and-databases/reference/query-set.md
@@ -262,10 +262,10 @@ query_set = Post.all
 query_set.prefetch(:author__favorite_tags)
 ```
 
-In some situations it might be necessary to use a custom query set for the prefetched records. This is possible by providing a `query_set` argument:
+In some situations, it might be necessary to use a custom query set for the prefetched records. This is possible by using a variant of the `#prefetch` method in which a single relation name and the associated query set (`query_set` argument) are provided:
 
 ```crystal
-# Query all lists and order the list items by its position
+# Query all lists and order the list items by position
 query_set = List.prefetch(:items, query_set: Item.order(:position))
 ```
 
--- a/docs/docs/models-and-databases/reference/query-set.md
+++ b/docs/docs/models-and-databases/reference/query-set.md
@@ -75,7 +75,7 @@ qset_2[2..6]? # returns a "sliced" query set
 
 Combines the current query set with another one using the **AND** operator.
 
-This method returns a new query set that is the result of combining the current query set with another one using the AND SQL operator. 
+This method returns a new query set that is the result of combining the current query set with another one using the AND SQL operator.
 
 For example:
 
@@ -90,7 +90,7 @@ combined_query_set = query_set_1 & query_set_2
 
 Combines the current query set with another one using the **OR** operator.
 
-This method returns a new query set that is the result of combining the current query set with another one using the OR SQL operator. 
+This method returns a new query set that is the result of combining the current query set with another one using the OR SQL operator.
 
 For example:
 
@@ -105,7 +105,7 @@ combined_query_set = query_set_1 | query_set_2
 
 Combines the current query set with another one using the **XOR** operator.
 
-This method returns a new query set that is the result of combining the current query set with another one using the XOR SQL operator. 
+This method returns a new query set that is the result of combining the current query set with another one using the XOR SQL operator.
 
 For example:
 
@@ -260,6 +260,13 @@ It should be noted that it is also possible to follow relations and reverse rela
 ```crystal
 query_set = Post.all
 query_set.prefetch(:author__favorite_tags)
+```
+
+In some situations it might be necessary to use a custom set for the prefetched records. This is possible by providing a `custom_query` argument:
+
+```crystal
+# Query all lists and order the list items by its position
+query_set = List.prefetch(:items, query_set: Item.order(:position))
 ```
 
 Finally, it is worth mentioning that multiple relations can be specified to `#prefetch`. For example:
--- a/docs/docs/models-and-databases/reference/query-set.md
+++ b/docs/docs/models-and-databases/reference/query-set.md
@@ -262,7 +262,7 @@ query_set = Post.all
 query_set.prefetch(:author__favorite_tags)
 ```
 
-In some situations it might be necessary to use a custom set for the prefetched records. This is possible by providing a `custom_query` argument:
+In some situations it might be necessary to use a custom query set for the prefetched records. This is possible by providing a `query_set` argument:
 
 ```crystal
 # Query all lists and order the list items by its position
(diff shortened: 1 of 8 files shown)


src/marten/db/query/prefetcher.cr
```diff
@@ -166,20 +171,32 @@ module Marten
...
         end
 
+        private def prefetch_query_set(relation_name, default_model)
+          if custom_query = @custom_query_sets[relation_name]?
+            raise Errors::MismatchedQuerySetType.new(
+              "Can't prefetch :#{relation_name} using #{custom_query.model} query set."
+            ) if custom_query.model != default_model
+            custom_query
+          else
+            default_model.unscoped
```
treagod, line 181: Just some thought:

Now that Marten provides the functionality to use any custom query, what do you think about reverting this to use the default scope?

For example if a `Board` has many users, which have a column `deleted_at`, it would be helpful to only load those that are not deleted.

I know this is a breaking change, but it makes more sense to me

ellmetha: For now I would say let's keep using `unscoped`. Your example makes sense when considering the case of soft-deletion, but default scopes can be overridden for things that are completely different. I don't know if/when we will add soft-deletion support to the framework, so in the meantime I don't think there is need for changing this default behavior.

src/marten/db/errors.cr
```diff
@@ -4,6 +4,9 @@ module Marten
       # Represents an error raised when a specific predicate will return an empty result set.
       class EmptyResults < Exception; end
 
+      # Represents an error raised when using incompatible query sets.
+      class MismatchedQuerySetType < Exception; end
```
ellmetha, line 8: I believe we can avoid introducing a new error type and we can use the existing `UnmetQuerySetCondition` exception class instead.

src/marten/db/query/set.cr
```diff
@@ -1474,8 +1516,12 @@ module Marten
           {% model_types = Marten::DB::Model.all_subclasses.reject(&.abstract?).map(&.name) %}
           {% if model_types.size > 0 %}
             alias Any = {% for t, i in model_types %}Set({{ t }}){% if i + 1 < model_types.size %} | {% end %}{% end %}
+            alias AnyQuerySet = {% for t, i in model_types %}
+              {{ t }}::QuerySet{% if i + 1 < model_types.size %} | {% end %}
```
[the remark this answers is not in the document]
treagod: What do you mean with **not used at all**? This is the union type of the `custom_query_sets`.

The compiler didn't accept any base class, so I had to resolve to the concrete class union. Do you have a better solution?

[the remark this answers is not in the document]
ellmetha: Having two similar aliases (`Set::Any` and `Set::AnyQuerySet`) is not ideal, we should avoid it if possible. Right now it appears that `Set::Any` is not used in the codebase and was never documented explicitly: maybe we can remove it so that we only have one alias (`Set::AnyQuerySet`)?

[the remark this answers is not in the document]
ellmetha: Additionally, I was also curious as to why it is necessary to introduce an additional alias for this. Why not use `Set::Any` directly?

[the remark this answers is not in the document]
treagod: Ah, got you!

Now you got me thinking. My first implementations didn't work with `Any`, as a result I tested it with a concrete `Type::QuerySet` class, which I later on extended to `AnyQuerySet`. I tested `Any` again now and I don't see any issues from my start implementation anymore. 

Removed the new alias and documented the new one. AFAIK this won't be considered by `crystal docs`?

src/marten/db/query/prefetcher.cr
```diff
@@ -166,20 +171,32 @@ module Marten
           records_to_decorate
         end
 
+        private def prefetch_query_set(relation_name, default_model)
+          if custom_query = @custom_query_sets[relation_name]?
+            raise Errors::UnmetQuerySetCondition.new(
+              "Can't prefetch :#{relation_name} using #{custom_query.model} query set."
```
ellmetha, line 177:
```suggestion
              "Cannot prefetch '#{relation_name}' relation using a #{custom_query.model.name} query set."
```

src/marten/db/query/prefetcher.cr
```diff
@@ -166,20 +171,32 @@ module Marten
           records_to_decorate
         end
 
+        private def prefetch_query_set(relation_name, default_model)
```
ellmetha, line 174: Can we rename this method to something like `#query_set_for_prefetched_relation`? 🙏

src/marten/db/query/set.cr
```diff
@@ -1179,6 +1187,40 @@ module Marten
...
+        # # "Can't prefetch :tags using Comment query set."
+        # ```
+        def prefetch(relation_name : String | Symbol, query_set : Any)
+          qs = clone
+          relation_name = relation_name.to_s
+          qs.prefetched_relations << relation_name
+
+          qs.custom_query_sets[relation_name] = query_set if query_set
+
+          qs
```
ellmetha, line 1221:
```suggestion
          relation_name = relation_name.to_s
          
          qs = clone
          qs.prefetched_relations << relation_name
          qs.custom_query_sets[relation_name] = query_set

          qs
```

(`query_set` will always be present)

src/marten/db/query/set.cr
```diff
@@ -1470,6 +1512,16 @@ module Marten
...
+
+        # Creates an alias `Any` representing the query sets of all existing models.
+        #
+        # For example, if you have the models `User` and `Post`,
+        # then this macro will generate:
+        #
+        #   alias Any = Set(User) | Set(Post)
+        #
+        # That means anywhere in your code where you accept `Any`, you accept
+        # any `Set(M)` specialized to these models.
```
ellmetha, line 1524: We don't need to document `finished` macros technically.

docs/docs/models-and-databases/reference/query-set.md
```diff
@@ -262,6 +262,13 @@ query_set = Post.all
 query_set.prefetch(:author__favorite_tags)
 ```
 
+In some situations it might be necessary to use a custom query set for the prefetched records. This is possible by providing a `query_set` argument:
```
ellmetha, line 265:
```suggestion
In some situations, it might be necessary to use a custom query set for the prefetched records. This is possible by using a variant of the `#prefetch` method in which a single relation name and the associated query set (`query_set` argument) are provided:
```

docs/docs/models-and-databases/reference/query-set.md
```diff
@@ -262,6 +262,13 @@ query_set = Post.all
...
+
+```crystal
+# Query all lists and order the list items by its position
```
ellmetha, line 268:
```suggestion
# Query all lists and order the list items by position
```

ellmetha (approved):
LGTM! Thanks for adding this! 🙏 
I mainly identified minor cosmetic things to improve before merging, see my comments below (and feel free to merge after updating the PR).

Issue #271, opened 2024-10-24
Improve Error Reporting for Route Parameter Mismatch


treagod:
# Description

Currently, when defining routes with Marten, route parameters need to align with the expected handler parameters. If there is a mismatch between the defined route parameter (e.g., `id`) and the expected parameter (e.g., `pk`), an error message like the following is raised:

```
'articles_show' route cannot receive {"pk" => 2} as parameters.
```

While the error is technically correct, it requires developers to stop and think about the mismatch, especially if the parameter names are similar or common (e.g., `id` vs. `pk`). A more descriptive error message could improve the developer experience by explicitly indicating both the expected and received parameters.

## Scenario

For example, in this route definition:

```crystal
path "/show/<id:int>", ArticlesShowHandler, name: "articles_show"
```

The handler `ArticlesShowHandler`,  expects a `id` parameter, but the reversed route uses `pk` instead. The resulting error is:

```
'articles_show' route cannot receive {"pk" => 2} as parameters.
```

The message could be more helpful by showing what parameter was expected versus what was provided.

#### Expected Parameters in Error Message:

Should the error message explicitly mention what parameters the handler expects, e.g.:

```
'articles_show' route cannot receive {"pk" => 2} as parameters. Expected parameters: ["id"].
```

#### Clarity for Parameter Mismatch:

How can we make this more intuitive for developers, especially in cases where common parameter names like `id` and `pk` are used?

### Benefits

- Improved Debugging: A clearer message would make it faster and easier to identify parameter mismatches.
- Better Developer Experience: Less cognitive overhead is required when interpreting error messages, especially for common parameter mismatches.

PR #290, opened 2025-02-01
Add more details to `NoReverseMatch` error 

treagod:
Adds more detail why a `#reverse` fails, e.g: 
`'home' route cannot receive {"sid" => "my-slug", "unknown_param" => "42"} as parameters. Missing: ["number"] Extra: ["unknown_param"]`

#271


ellmetha (commented):
Thanks for kickstarting this!

I have a few concerns with this change. It is definitely good to optimize for developer experience and try to provide better error messages, but we must do it in a way that preserves performance.

There are a few things that I am not aligned with here:

* We are optimizing for the exceptional case (ie. bad reverse uses) _by default_. We are adding validations that are executed upfront by assuming that parameters might be errored instead of _reacting_ to the presence of invalid parameters _lazily_ (and assuming the happy path by default).
* We are introducing unnecessary abstractions that are used systematically (`ReverseResult`) even though these improved messages should probably only be a thing that interests us in debug situations.

I ran a quick benchmark and it turns out that the new implementation is slower than the previous one:

```
Old reverse   3.44M (290.98ns) (± 0.95%)    912B/op        fastest
New reverse   2.68M (373.78ns) (± 2.46%)  1.14kB/op   1.28× slower
```

Given that we are introducing this slowness just for the sake of generating better exception messages, I am a bit on the fence on approving this change.

Could we find a way to remain closer to the original implementation and end up generating more complicated error message **only** when we are sure that the parameters are not the good ones (and avoid unnecessary "result" abstractions)? 🙏

ellmetha (approved):

PR #292, opened 2025-02-04
Fix `blank: true` description

whtsht:
While reading [Field validation rules](https://martenframework.com/docs/models-and-databases/validations#field-validation-rules), I found a statement that appears to be incorrect.

- fields using `blank: true` won't accept empty values

This contradicts the explanation in [fields#blank](https://martenframework.com/docs/models-and-databases/reference/fields#blank).

The correct explanation should be either
- fields using `blank: true` accept empty values
- fields using `blank: false` won't accept empty values

The latter is described earlier in this document, so I have fixed the explanation for `blank: true`.


whtsht:
I’m not sure why the CI is failing — any insights would be appreciated.

treagod:
> I’m not sure why the CI is failing... any insights would be appreciated!

Hi @whtsht,

Ameba's `master` was updated just before your PR, which checks the indentation of heredocs.

As marten requires the `master` of Ameba the QA is failing.

I'm going to address these issues and after a rebase everything should be ok with your PR

PR #302, opened 2025-05-01
query: Add total_count method to support pagination

miry:
Adds a `total_count` method that returns the total number of records in the queryset without applying pagination.

This is useful for paginators that need both the total number of pages and the total record count. Instead of recalculating the total size repeatedly, this method centralizes and reuses that logic.

Inspired by [kaminari](https://github.com/kaminari/kaminari/blob/ca4a5dcfce40ede7990ebfe00a12f21e78e910d9/kaminari-activerecord/lib/kaminari/activerecord/active_record_relation_methods.rb#L17)

### DIFF
--- a/spec/marten/db/query/paginator_spec.cr
+++ b/spec/marten/db/query/paginator_spec.cr
@@ -144,7 +144,7 @@ describe Marten::DB::Query::Paginator do
       paginator.total_count.should eq 2
     end
 
-    it "returns the expected pages count if there are multiple pages" do
+    it "returns the total count if there are multiple pages" do
       Tag.create!(name: "a_tag", is_active: true)
       Tag.create!(name: "b_tag", is_active: true)
       Tag.create!(name: "c_tag", is_active: true)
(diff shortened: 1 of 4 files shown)


spec/marten/db/query/paginator_spec.cr
```diff
@@ -129,4 +129,30 @@ describe Marten::DB::Query::Paginator do
...
+
+    it "returns the expected pages count if there are multiple pages" do
+      Tag.create!(name: "a_tag", is_active: true)
+      Tag.create!(name: "b_tag", is_active: true)
+      Tag.create!(name: "c_tag", is_active: true)
+      Tag.create!(name: "d_tag", is_active: true)
+      Tag.create!(name: "e_tag", is_active: true)
+
+      paginator = Tag.all.order(:name).paginator(2)
+      paginator.pages_count.should eq 3
```
ellmetha, line 155: This line should be as follows, since we are testing the `#total_count` method:

```suggestion
      paginator.total_count.should eq 5
```

ellmetha (commented):
Changes look good! Thank you for adding this!
Just one minor update to make in one of the specs, and then I'll happily merge this. 🙂

spec/marten/db/query/paginator_spec.cr
miry: good catch. fixed in [28282c8](https://github.com/martenframework/marten/pull/302/commits/28282c8c64b0e6ad776eb72c56cab15953082579)

ellmetha (approved):

PR #306, opened 2025-05-31
Fix `ManyToManySet#add` if an existing relation has no `id` field

treagod:
If a related m2m model has no `id` field as pk and it is added to the m2m relation twice following error is raised:

```bash
Error: undefined method 'id' for Country
```


treagod:
To be honest, I don't understand why the specs are failing.. is this an CI issue?

ellmetha (approved):
Nice catch! Thanks for the fix!

(I'll look into the CI issue separately)

PR #313, opened 2025-07-11
fix(media): support URL‑encoded filenames in development file lookup

treagod:
Previously, files with percent‑encoded characters (e.g. "%20") in media paths were not found. Now decode the requested path before lookup, ensuring media assets with spaces or other URL‑encoded names are served.


ellmetha (approved):
Thanks for fixing this!

PR #322, opened 2025-09-08
Feature/dev live reload

johnjansen:


ellmetha:
Hey! 👋 

Thanks a lot for opening this PR! I agree that live reloading could be useful for some Marten projects. However, I’m not sure it’s something that should live within the framework itself — many projects can (and do) use external tools for live- or hot-reloading, and the need might not be common enough to justify maintaining it as part of Marten.

I think it would make a great standalone shard though. Would you be open to creating one?

johnjansen:
**[jj.monitor]** Branch is 106 commits behind `main` and has substantive conflicts — 12 files diverged including `src/marten/server.cr`, `src/marten/conf/global_settings.cr`, and several middleware/handler files that have likely been rewritten upstream since Sep 2025.

The 14 failing CI checks (crystal, generated-project, postgres matrix, generated-auth-project) are downstream of the staleness, not all independent issues — they need re-running against a freshly-rebased branch before they're meaningful signal. Recommend rebasing onto current `main` in a worktree and pushing, then triaging whatever still fails.

`[ƒ:kx7]`

PR #323, opened 2025-09-17
Use EnvironmentFiles to load settings and add User to Systemd service file

bitmand:
Two documentation changes:

- add `User=deploy` to systemd service file, so the service runs as the deploy user (as mentioned earlier on the same page).
- recommend using `EnvironmentFiles` to load a `settings.env` for secrets and sensitive settings.

Having secrets/sensitive information in Systemd service files is not a good idea, since any user on the same host can read it - eg. try:

    systemctl show <yourapp>.service | grep -i Environment


ellmetha (approved):
Makes sense - thanks for updating this!

PR #329, opened 2025-11-09
Demonstrate issue with JSON payload and Schema fields parsing

miry:
## Summary

This PR introduces the use of JSON payloads and highlights an issue with how JSON data is currently parsed in `Marten::HTTP::Request` for `#data`.

While testing JSON payload parsing, I noticed that all root-level keys in the JSON payload are being converted to arrays. [src](https://github.com/martenframework/marten/blob/a72089b383d3ad7ce424d81b4547de47685aa109/src/marten/http/request.cr#L343).

## Problems

1. The request.data hash generated from a JSON payload does not match the original JSON structure.

2. While this behavior doesn’t affect `string` fields when using the Schema feature, it breaks for `array` fields, which are not parsed correctly. Example: `{"foo": ["a", "b"]}` would treated in Schema as `foo => '["a","b"]'`. and schema:
```
  class RequestSchema < Marten::Schema
    field :foo, :array, of: :string, max_size: 2, required: true
  end

  schema = RequestShcema.new(request.data)

  schema.foo # => '["a", "b"]'
```

PS: This PR contains changes from https://github.com/martenframework/marten/pull/328

### DIFF
--- a/spec/marten/spec/client_spec.cr
+++ b/spec/marten/spec/client_spec.cr
@@ -230,12 +230,13 @@ describe Marten::Spec::Client do
       client = Marten::Spec::Client.new
 
       response = client.patch(
-        Marten.routes.reverse("simple_schema"),
+        Marten.routes.reverse("request_data_respond"),
         data: {first_name: "John", last_name: "Doe"}.to_json,
         content_type: "application/json"
       )
 
-      response.status.should eq 302
+      response.status.should eq 200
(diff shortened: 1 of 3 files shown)


ellmetha:
Hello! 👋 This should be fixed by [9fd3078](https://github.com/martenframework/marten/commit/9fd3078cd2fa2908c814c681d9f04eb6e4c017a8).

PR #335, opened 2025-12-27
Add update_columns methods to DB::Model for partial updates

miry:
## Problem

I found that if it is only require to update a single column in DB, all other fields also send to DB. Example of Logs for changing only field `public_profile`:

```
Executing query: UPDATE "users" SET "created_at"=$1, "updated_at"=$2, "email"=$3, "name"=$4, "code"=$5, "nickname"=$6, "public_profile"=$7, "referral_code"=$8, "referrer_id"=$9, "official_tag_en"=$10, "official_tag_uk"=$11, "total_filament_weight_printed"=$12, "total_filament_weight_printed_last_30"=$13 WHERE "id"=$14 -- prefix: "  ▸ "
```

## Solution

Introduce new methods `update_columns` and `update_columns!` that allow updating specific database columns without running validations or callbacks. These methods provide a performance optimization for cases where only a subset of fields needs to be updated.

Key features:
- Updates only specified columns, leaving others unchanged
- Bypasses model validations and lifecycle callbacks
- Updates both in-memory values and database records
- `update_columns!` raises an error for unsaved records

Usage:
  user.update_columns(username: "new_name") user.update_columns!(email: "new@example.com", last_login: Time.utc)

It generates logs:

```
Executing query: UPDATE "users" SET "public_profile"=$1 WHERE "id"=$2
```

### DIFF
--- a/spec/marten/db/model/persistence_spec.cr
+++ b/spec/marten/db/model/persistence_spec.cr
@@ -1579,6 +1579,17 @@ describe Marten::DB::Model::Persistence do
       object.username.should eq "test1"
       object.email.should eq "test1@example.com"
     end
+
+    it "does not allow to update for a new record" do
+      object = TestUser.new(first_name: "John", last_name: "Doe")
+      object.update_columns({"username" => "test1", "email" => "test1@example.com"}).should be_false
+    end
+
+    it "does not allow to update for a destroyed record" do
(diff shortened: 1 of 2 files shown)


ellmetha (commented):
I have one question, but the changes look good otherwise!

ellmetha (approved):
Thanks for adding this!

PR #337, opened 2026-02-02
Implement `update_or_create` method

### LINKED ISSUE #308: Add an `#update_or_create` method to query sets
# Description

Let's introduce a new `#update_or_create` method to query sets.

## Rationale

Let's make it easy to either update an existing record or create a new one with a set of specified attributes through the use of a new `#update_or_create` query set method. A simple use of this method would look something like this:

```crystal
Person.all.update_or_create(updates: { first_name: "Bob" }, first_name: "John", last_name: "Doe")
```

By default, the set of attributes specified in the `updates` argument should be used to update the existing record or initialize the newly created record.

Alternatively, it should be possible to specify an alternative `defaults` argument that should be used instead of the `updates` one for initializing newly created records (this is important if additional attributes must be set on new records but are not necessary for updated records):

```crystal
Person.all.update_or_create(
  updates: { first_name: "Bob" }, 
  defaults: { first_name: "Bob", active: true }, 
  first_name: "John", 
  last_name: "Doe"
)
```

treagod:
This commit adds a new `#update_or_create` method to the querysets.

With this newly added method a developer can update specific fields, defined by `updates`, of record that matches the filter or, if no record can be found, create a new record. The new record will either created by the `updates` arguments or a custom `defaults` hash can be used if creation requires certain fields to be set

Closes #308

### DIFF
--- a/spec/marten/db/model/querying_spec.cr
+++ b/spec/marten/db/model/querying_spec.cr
@@ -1745,7 +1745,7 @@ describe Marten::DB::Model::Querying do
       Marten::DB::Model::QueryingSpec::Tag.all.size.should eq 2
     end
 
-    it "uses defaults when creating a new record if they are provided" do
+    it "uses updates even when defaults are specified when creating a new record if they are provided" do
       new_tag = Marten::DB::Model::QueryingSpec::Tag.update_or_create(
         updates: {name: "unused", is_active: true},
         defaults: {name: "newtag", is_active: false},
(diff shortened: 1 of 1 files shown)


src/marten/db/query/set.cr
```diff
@@ -1577,6 +1577,46 @@ module Marten
...
+        # exception if multiple records match the specified set of filters.
+        def update_or_create(
+          *,
+          updates : Hash | NamedTuple,
+          defaults : Hash | NamedTuple | Nil = nil,
+          **kwargs,
+        )
+          record = get!(Node.new(**kwargs))
+          record.set_field_values(updates)
+          record.save(using: @query.using)
```
ellmetha: I guess it could be interesting to have an `#update_or_create!` variant of the method where we call `#save!` so that an exception is raised if the created/updated record is invalid.

PR #341, opened 2026-03-06
Query set enum input

treagod:
This PR adds the possibility to use an enum value to set a field value

### DIFF
--- a/src/marten/db/field/enum.cr
+++ b/src/marten/db/field/enum.cr
@@ -4,17 +4,14 @@ module Marten
       # Represents an enum field.
       class Enum < Base
         @default : ::String?
-        @enum_type : ::Enum.class | Nil
 
         getter default
-        getter enum_type
-
-        getter values
+        getter enum_class
 
         def initialize(
           @id : ::String,
-          enum_values : Array(::String),
-          @enum_type : ::Enum.class | Nil = nil,
+          *,
+          @enum_class : ::Enum.class,
           @primary_key = false,
           @blank = false,
           @null = false,
@@ -24,8 +21,11 @@ module Marten
           default : ::Enum? = nil,
           **kwargs,
         )
-          @values = enum_values
           @default = default.try(&.to_s)
+        end
+
+        def values : Array(::String)
+          enum_class.names
         end
 
         def from_db(value) : ::String?
@@ -62,7 +62,7 @@ module Marten
           when Symbol
             value.to_s
           when ::Enum
-            if !enum_type.nil? && value.class != enum_type
+            if value.class != enum_class
               raise_unexpected_field_value(value)
             end
             value.to_s
@@ -86,13 +86,12 @@ module Marten
 
           class ::{{ model_klass }}
             register_field(
-                {{ @type }}.new(
-                  {{ field_id.stringify }},
-                  {% unless kwargs.is_a?(NilLiteral) %}**{{ kwargs }}{% end %},
-                  enum_type: {{ enum_klass }},
-                  enum_values: {{ enum_klass }}.values.map(&.to_s),
-                )
+              {{ @type }}.new(
+                {{ field_id.stringify }},
+                {% unless kwargs.is_a?(NilLiteral) %}**{{ kwargs }}{% end %},
+                enum_class: {{ enum_klass }},
               )
+            )
 
             {% if !model_klass.resolve.abstract? %}
               @[Marten::DB::Model::Table::FieldInstanceVariable(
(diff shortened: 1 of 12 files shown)


src/marten/db/field/enum.cr
```diff
@@ -4,14 +4,17 @@ module Marten
...
 
         getter default
+        getter enum_type
 
         getter values
 
         def initialize(
           @id : ::String,
           enum_values : Array(::String),
+          @enum_type : ::Enum.class | Nil = nil,
```
ellmetha: Why is this nilable? Technically, it shouldn't be possible to have an enum field without an actual enum class.

src/marten/db/model/table.cr
```diff
@@ -615,8 +620,9 @@ module Marten
         def set_field_values(values : Hash | NamedTuple) # ameba:disable Naming/AccessorMethodName
           sanitized_values = Hash(String, Field::Any | Model).new
           values.each do |key, value|
-            next unless value.is_a?(Field::Any | Model)
-            sanitized_values[key.to_s] = value
+            accepted, normalized_value = normalize_set_field_value(value)
```
ellmetha: Can we avoid using method overloading for this? This makes the code more difficult to maintain for very little gains I believe. I standard case/when would work perfectly fine here:

```ruby
          values.each do |key, value|
            sanitized_value = case value
            when Field::Any, Model
              value
            when ::Enum
              value.to_s
            else
              next
            end

            sanitized_values[key.to_s] = value
          end
```

src/marten/db/query/set.cr
```diff
@@ -1550,7 +1550,18 @@ module Marten
...
+            when Field::Any, DB::Model
+              update_hash[k.to_s] = v
+            when ::Enum
+              update_hash[k.to_s] = normalize_enum_update_value(k, v)
+            else
+              raise Errors::UnexpectedFieldValue.new(
+                "Value '#{v.inspect}' cannot be used in update queries"
+              )
+            end
+          end
```
ellmetha: I am not convinced that we should introduce this type of logic at this level. The `Query#update_with` method already calls `#to_db` for each of the fields corresponding to the values specified as input. I would argue that we should try to reuse this as much as possible instead of replicating some of this logic at the set level (whose responsibility is mainly to enable the query set interface and to leave everything related to data preparation to the query abstraction).

src/marten/db/query/set.cr
```diff
@@ -1679,6 +1690,17 @@ module Marten
...
 
+        private def normalize_enum_update_value(field_name : String | Symbol, value : ::Enum) : Field::Any
+          field = begin
+            M.get_field(field_name)
+          rescue Errors::UnknownField
+            nil
+          end
+          return value.to_s if field.nil? || !field.is_a?(Field::Enum)
+
+          field.as(Field::Enum).to_db(value).as(Field::Any)
```
ellmetha: Additionally, outside of relation fields, we should avoid making assumptions regarding the type of fields being manipulated. As such, we should not explicitly check that the manipulated field is an enum field, ideally.

src/marten/db/query/sql/query.cr
```diff
@@ -458,8 +458,21 @@ module Marten
...
+          def update_with(values : Hash(String | Symbol, Field::Any | DB::Model | EnumType)) forall EnumType
+            {% unless EnumType <= ::Enum %}
+              {% raise "Query#update_with only accepts field values, model instances, and enum values" %}
+            {% end %}
+
+            internal_update_with(values.transform_keys(&.to_s))
+          end
+
+          # :nodoc:
+          def internal_update_with(values : Hash(String, V)) forall V
```
ellmetha: I am really not convinced by the need for adding this, to be honest. This complexifies the implementation unnecessarily, and this is not consistent with most of the query set methods in any case.

Can't we have something working without explicitly expecting enums? It looks to me that a very simple change like this is all we need (no need to introduce macros for that!):

```diff
diff --git a/src/marten/db/query/set.cr b/src/marten/db/query/set.cr
index 25e153b7..aa0117cf 100644
--- a/src/marten/db/query/set.cr
+++ b/src/marten/db/query/set.cr
@@ -1550,7 +1550,14 @@ module Marten
         # either.
         def update(values : Hash | NamedTuple)
           update_hash = Hash(String | Symbol, Field::Any | DB::Model).new
-          update_hash.merge!(values.to_h)
+          values.each do |key, value|
+            case value
+            when Field::Any, DB::Model
+              update_hash[key.to_s] = value
+            else
+              update_hash[key.to_s] = value.to_s
+            end
+          end

           qs = clone
           updated_count = qs.query.update_with(update_hash)
```

ellmetha (approved):

PR #343, opened 2026-03-20
Add email attachment support

### LINKED ISSUE #320: Add support for email attachments
# Description

It is presently not possible to associate attachments with emails when using [emailing capabilities](https://martenframework.com/docs/emailing). Let's add support for this.

## Proposition

In order to add attachments to an email, it should be possible to leverage a new `#attach` method. This method should allow associating a file (path, actual `File` object, or `IO::Memory` object) with the considered email, while also allowing specifying the attachment name and mime type if necessary.

The new `#attach` method should be called as part of a `before_deliver` callback. For example:

```crystal
class WelcomeEmail < Marten::Email
  from "no-reply@martenframework.com"
  to @user.email
  subject "Hello!"
  template_name "emails/welcome_email.html"

  before_deliver :add_attachments

  def initialize(@user : User)
  end

  private def add_attachments
    # Add an attachment from a static path
    attach "public/docs/terms.pdf"
    
    # Add an attachment from a generated IO
    attach generate_pdf_io(@user), filename: "welcome.pdf", mime_type: "application/pdf"
  end
end
```

treagod:
This PR introduces e-mail attachment support to Marten's emailing system.

It adds a attachment struct and new `attach` helpers for file paths, `File` instances, and generic `IO` objects. MIME types are inferred from the filename when possible, with a fallback to `application/octet-stream`.

The change also makes attachments accessible to other backends and includes attachment details in the development backend output.

Closes #320

### DIFF
--- a/docs/docs/emailing/callbacks.md
+++ b/docs/docs/emailing/callbacks.md
@@ -60,6 +60,8 @@ class WelcomeEmail < Marten::Email
 end
 ```
 
+This callback is also a common place to call [`#attach`](./introduction.md#defining-attachments) in order to add email attachments right before delivery.
+
 ### `after_deliver`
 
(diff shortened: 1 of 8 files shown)


ellmetha (approved):
LGTM!

PR #353, opened 2026-04-09
Update path for `bin/ameba.cr` executable in `lint` Makefile target

Sija:
Refs https://github.com/crystal-ameba/ameba/pull/807


ellmetha (approved):
Thanks!

PR #354, opened 2026-04-12
feat: allow hash keys to be resolved as template attributes

papilip:
## Summary

Allow `Hash` keys to be used as template attributes, so that `{{ my_hash.my_key }}` resolves to the value associated with `my_key`.

**Before:** Only methods listed in `template_attributes` (like `empty?`, `size`, `keys`) were accessible. Hash keys were silently ignored, resulting in empty values with no error — a common pitfall.

**After:** Hash keys are resolved first, with a fallback to the standard `template_attributes` methods.

## Motivation

When passing a `Hash(String, String)` to a template context:

```crystal
render "my_template.html", context: {
  item: {"type" => "order", "amount" => "60.00"},
}
```

Users expect `{{ item.type }}` to output `order`. Instead, it silently outputs nothing because `Hash` only exposes methods like `empty?` and `size` to the template engine.

This is particularly confusing because:
- No error is raised — values are just empty
- `NamedTuple` works as expected via `Object::Auto`
- The fix is non-obvious (requires creating a custom struct with `Object::Auto`)

## Changes

- **`src/marten/template/ext/hash.cr`**: Override `resolve_template_attribute` to check `has_key?(key)` before falling back to the macro-generated method
- **`spec/marten/template/ext/hash_spec.cr`**: 4 new tests covering key access, fallback to standard attributes, and error on unknown keys

## Test plan

- [x] Existing hash spec tests still pass (9 tests)
- [x] New tests for key access (4 tests)
- [x] `size` attribute still works when a key named `size` does not exist (fallback)
- [x] `UnknownVariable` is raised for non-existent keys that are not standard attributes

### DIFF
--- a/src/marten/template/ext/hash.cr
+++ b/src/marten/template/ext/hash.cr
@@ -3,4 +3,16 @@ class Hash
   include Marten::Template::CanDefineTemplateAttributes
 
   template_attributes :any?, :compact, :empty?, :keys, :none?, :one?, :present?, :size, :values
+
+  # Allows to resolve hash keys as template attributes.
+  #
+  # This makes it possible to access hash values using `{{ my_hash.my_key }}` in templates,
+  # in addition to the standard hash methods (empty?, size, keys, etc.).
+  def resolve_template_attribute(key : ::String)
+    if has_key?(key)
+      self[key]
+    else
+      previous_def
+    end
+  end
 end
--- a/spec/marten/template/ext/hash_spec.cr
+++ b/spec/marten/template/ext/hash_spec.cr
@@ -46,5 +46,24 @@ describe Hash do
     it "returns the expected result when requesting the 'values' attribute" do
       ({"a" => 1, "b" => 2, "c" => 3} of String => Int32).resolve_template_attribute("values").should eq [1, 2, 3]
     end
+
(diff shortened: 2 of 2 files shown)


src/marten/template/ext/hash.cr
```diff
@@ -3,4 +3,16 @@ class Hash
...
+  # Allows to resolve hash keys as template attributes.
+  #
+  # This makes it possible to access hash values using `{{ my_hash.my_key }}` in templates,
+  # in addition to the standard hash methods (empty?, size, keys, etc.).
+  def resolve_template_attribute(key : ::String)
+    if has_key?(key)
+      self[key]
+    else
+      previous_def
+    end
```
ellmetha, line 16: Unfortunately, this change has serious limitations. If a hash contains a key like `size`, then it will be impossible to get the size from the hash when calling `{{ my_hash.size }}`. Similarly, any other method specified in `#template_attributes` will only work as long as a key with the same name is not part of the hash.

I think that we should give access to hash methods with priority over keys. That said, I don't want to allow hash key lookups if there is not a built-in way to actually do key lookups for keys such as "size".

The real way forward would be to add both support for dotted access and bracket access (eg. `{{ my_hash["size"] }}`).

Issue #165, opened 2024-02-18
Add a Component template tag


treagod:
# Description

Let's enhance Marten's templating system by introducing a new component tag. Unlike the existing include tag, the component tag provides a more structured and robust approach for building reusable components within templates.

## Background

While the include tag serves its purpose for certain scenarios, it falls short when it comes to building complex applications with reusable components. There is a need for a more sophisticated mechanism that ensures better encapsulation, parameterization, and error handling within components.

## Proposed Solution

Introduce a component tag to Marten's templating system, which offers the following key features:

- Components cannot access variables in the global template scope; they only have access to variables provided by context producers.
- Components must specify which parameters they require, indicating whether a parameter is required or optional with a default value.
- Instantiating a component without a required parameter would result in an error, providing better error handling and preventing runtime issues.
- Components can define slots, allowing flexibility in content insertion and customization within the component. Slots shouldn't have access to the components scope, but to the outer scope.

## Functionality

- The component tag will be used to define reusable components within Marten templates.
- Components will declare their required parameters and optional parameters with default values, ensuring clear communication of dependencies.
- When instantiating a component, the template engine will enforce parameter requirements and handle errors gracefully if any required parameters are missing.
- Components can define slots within their markup, allowing dynamic content insertion and customization.

Example Usage:

```html
{# Define component  #}
{% define_component "button" with type: String %}
<button class="{{ type }}%>
  {% slot %}Default slot text{% endslot %}
</button>
{% enddefine_component %}

{# Use component  #}
{% component "button" with type: "primary" %}
  Slot Text
{% endcomponent %}
```

## Discussion

Where are components stored? How are they loaded when they are referenced?

ellmetha:
Is this mechanism similar to [Jinja macros](https://jinja.palletsprojects.com/en/3.1.x/templates/#macros)?

Considering this example, I'm a bit uncertain about the value of this mechanism. As far as I can see, this example could be implemented very easily using a snippet and the [`include`](https://martenframework.com/docs/templates/reference/tags#include) template tag.

For example, the snippet could look something like this:

```html
<button class="{{ type }}">
  {{ text | default: "Default slot text" }}
</button>
```

And then you could use this snippet as follows:

```liquid
{% include "my_snippet" with type="primary" %}
{% include "my_snippet" with type="primary", text="Slot text" %}
```

Why couldn't the `include` template tag be sufficient for this or similar use cases? Are there other use cases where `include` is clearly not sufficient?

[...]

PR #355, opened 2026-04-12
feat: add render_partial for HTMX fragment responses

papilip:
## Summary

Add a `render_partial` method to `Marten::Handlers::Base` that renders a template **without** the layout, context producers, or `before_render` callbacks.

## Motivation

HTMX is increasingly popular for building interactive web applications. A common pattern is returning HTML fragments from handlers to replace parts of the page:

```crystal
def post
  # Update something...
  render_partial("shared/_item_card.html", context: {item: item})
end
```

Currently, developers must work around this by manually loading and rendering templates:

```crystal
tpl = Marten.templates.get_template("shared/_item_card.html")
ctx = Marten::Template::Context{"item" => item}
Marten::HTTP::Response.new(tpl.render(ctx), content_type: "text/html", status: 200)
```

This is verbose, error-prone, and bypasses the handler's response infrastructure.

## Changes

- **`src/marten/handlers/base.cr`**: New `render_partial` method that creates a clean `Context` (no producers), renders the template, and returns an `HTTP::Response`
- **`spec/marten/handlers/base_spec.cr`**: 3 tests (hash context, named tuple context, custom status)

## Design decisions

- **No context producers**: Partials are fragments — they should not include layout-level data
- **No `before_render` callbacks**: These are designed for full-page renders
- **No `request` in context**: Keeps the partial isolated. If needed, the caller can add it explicitly
- Supports `Hash`, `NamedTuple`, `Context`, and `nil`

## Related

- Issue #165 (Component template tag) — `render_partial` addresses the server-side use case

## Test plan

- [x] Renders template with hash context
- [x] Renders template with named tuple context  
- [x] Custom status code works
- [x] All 71 existing handler base tests still pass

### DIFF
--- a/spec/marten/handlers/base_spec.cr
+++ b/spec/marten/handlers/base_spec.cr
@@ -1146,6 +1146,55 @@ describe Marten::Handlers::Base do
       handler.url("dummy_with_id", id: 42).should eq Marten.routes.reverse("dummy_with_id", id: 42)
     end
   end
+
+  describe "#render_partial" do
+    it "renders a template without context producers or layout" do
+      request = Marten::HTTP::Request.new(
+        ::HTTP::Request.new(
+          method: "GET",
+          resource: "",
+          headers: HTTP::Headers{"Host" => "example.com"}
+        )
+      )
+
+      handler = Marten::Handlers::BaseSpec::Test5Handler.new(request)
+      response = handler.render_partial("specs/handlers/base/test.html", context: {"name" => "Partial"})
+
+      response.status.should eq 200
+      response.content.strip.should eq "Hello World, Partial!"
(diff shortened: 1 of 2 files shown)


ellmetha (commented):
I am not aligned with the idea of introducing a method that essentially allows skipping `before_render` callbacks and context producers. In Marten’s handler design, `#render` already covers fragment responses for most cases (there is no concept of layout in Marten!), and bypassing lifecycle hooks introduces two rendering pipelines with different invariants, which is surprising and easy to misuse.

Since there is no concept of layout in Marten, `#render` can already be used to render "partials". So why not use it directly? I don't think that having context producers being used is that detrimental here.

PR #364, opened 2026-05-13
Preserve `prefetched_relations` through `RelatedSet` and `ManyToManySet` clones

stevegeek:
### Background: 

`Marten::DB::Query::Set#prefetch` returns a cloned queryset whose `prefetched_relations` collection is then updated in place. For the top-level `Set` this works because `Set#clone` forwards `prefetched_relations` (and `custom_query_sets`) to the new instance, so any subsequent chained method (e.g. `.order`, `.filter`, `.exclude`) carries the prefetch state forward and `#fetch` ultimately runs the prefetcher.

### Issue:

`RelatedSet` and `ManyToManySet` are different in two ways:

1. Their `#initialize` overrides do not accept `prefetched_relations` or `custom_query_sets` kwargs (the parent `Set#initialize` does), so newly-constructed instances always start with empty state.

2. Their `#clone` overrides construct a new instance without forwarding either field.

The result is that calling `#prefetch` on a `RelatedSet` (e.g. via a reverse many-to-one accessor like `user.posts`) appears to work but any chained method that calls `#clone` strips it. 

The next call to `#fetch` finds an empty `prefetched_relations`, skips the prefetcher entirely, and relation accessors on each returned record fall back to N+1 

```crystal
# Each post.author triggers a SELECT because the .order clone
# stripped the prefetch.
user.posts.prefetch(:author).order(:created_at).each { |p| p.author }
```

The macro-generated `<Model>::RelatedQuerySet` and `<Model>::ManyToManyQuerySet` in `model/querying.cr` have the same pair of overrides and inherit the bug.

### Fix:

This commit threads both fields through `#initialize` and `#clone` on:

- `Marten::DB::Query::RelatedSet`
- `Marten::DB::Query::ManyToManySet`
- `<Model>::RelatedQuerySet` (generated)
- `<Model>::ManyToManyQuerySet` (generated)

so prefetch state survives regardless of chain order.

### DIFF
--- a/spec/marten/db/query/related_set_spec.cr
+++ b/spec/marten/db/query/related_set_spec.cr
@@ -159,4 +159,32 @@ describe Marten::DB::Query::RelatedSet do
       qset[0].get_related_object_variable(:author).should be_nil
     end
   end
+
+  describe "#prefetch with chained operations" do
+    it "preserves prefetched_relations through #order" do
+      user_1 = TestUser.create!(username: "jd1", email: "jd1@example.com", first_name: "John", last_name: "Doe")
+      user_2 = TestUser.create!(username: "jd2", email: "jd2@example.com", first_name: "John", last_name: "Doe")
+
+      post_1 = Post.create!(author: user_1, title: "Post 1")
+      Post.create!(author: user_2, title: "Post 2")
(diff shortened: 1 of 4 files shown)


ellmetha (approved):
Looks good. Thanks for the fix!

PR #378, opened 2026-05-21
Add `--grep` option to `marten routes`

### LINKED ISSUE #262: Add Filtering Option to the `marten routes` Command
# Description

Currently, the `marten routes` command lists all the routes of the application. To enhance usability, especially in larger applications with numerous routes, it would be beneficial to introduce an option that allows users to filter routes based on a substring (or regex). This option would enable developers to quickly locate routes that partially match a specified pattern, either in the URL path or the URL helper method.

## Proposition

Add a `-g` option the `marten routes` command. For example running `marten routes -g admin` should output only those routes where “admin” appears in the URL path or the helper method.

# Benefits:

- **Improved Usability**: Makes it easier for developers to find specific routes in large applications.
- **Efficiency**: Reduces the time spent manually searching through routes, especially in complex projects.

treagod:
Closes #262

### DIFF
--- a/docs/docs/development/reference/management-commands.md
+++ b/docs/docs/development/reference/management-commands.md
@@ -231,6 +231,10 @@ marten resetmigrations foo # Resets the migrations of the "foo" application
 
 Displays all the routes of the application.
 
+### Options
+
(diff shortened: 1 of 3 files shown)


ellmetha (approved):
Looks good. Thanks for adding this!

PR #414, opened 2026-08-11
Add support for raw update queries

### LINKED ISSUE #390: Add support for raw update queries
# Description

Let's add support for raw update queries.

## Rationale

Marten presently supports [updating records](https://martenframework.com/docs/models-and-databases/reference/query-set#update) through query sets using the [`#update`](https://martenframework.com/docs/models-and-databases/reference/query-set#update) method, but this method only accepts a hash of field names and literal values. It is not possible to perform updates where the new value is computed at the SQL level (eg. copying one column into another).

[Raw SQL predicates](https://martenframework.com/docs/models-and-databases/raw-sql#filtering-with-raw-sql-predicates) are already supported for filtering query sets. Let's extend a similar approach to updates.

### Basic usage

```crystal
Invoice.filter(number__isnull: true).update("number = id")
```

The passed string should be used as the `SET` clause of the produced `UPDATE` statement. Column names should refer to database column names.

### With parameters

Positional (`?`) and named (`:param`) parameters should be supported, similarly to raw SQL predicates:

```crystal
Invoice.filter(status: "draft").update("label = ?", "default")
Invoice.filter(status: "draft").update("label = :label", label: "default")
```

### Restrictions

* Raw updates should follow the same semantics as regular `#update` calls: no validation and no callbacks are executed.
* Developers are responsible for ensuring that the specified SQL is valid and safe (parameters should always be used instead of string interpolations).

w3lld1:
## Summary

- I added raw `SET` expression overloads to query set updates.
- I added positional and named parameter support while preserving parameter offsets for filtered updates.
- I added regression coverage for direct expressions, both parameter styles, and empty expressions.

## Validation

- `crystal spec ./spec/marten/db/query/set_spec.cr --error-trace` — 325 examples, 0 failures
- `make qa_crystal` — formatter passed; 1,328 files inspected by Ameba with 0 failures
- `git diff --check`

I also started the full `make tests` suite locally, but it exceeded the 10-minute local execution limit. The focused query set suite and the complete Crystal QA checks passed; the full database matrix is deferred to CI.

Fixes #390

### DIFF
--- a/spec/marten/db/query/set_spec.cr
+++ b/spec/marten/db/query/set_spec.cr
@@ -3678,6 +3678,57 @@ describe Marten::DB::Query::Set do
       user_3.is_admin.should be_falsey
     end
 
+    it "allows to update records using a raw SQL expression" do
+      user = TestUser.create!(username: "abc", email: "abc@example.com", first_name: "John", last_name: "Doe")
+
+      qset = Marten::DB::Query::Set(TestUser).new
+
+      qset.filter(username: "abc").update("first_name = last_name").should eq 1
+
+      user.reload
+      user.first_name.should eq "Doe"
+    end
+
+    it "allows to use positional parameters in raw SQL update expressions" do
+      user = TestUser.create!(
+        username: "abc",
+        email: "abc@example.com",
+        first_name: "John",
+        last_name: "Doe"
+      )
+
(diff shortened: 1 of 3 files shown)


ellmetha (approved):
Thanks for adding this!

PR #428, opened 2026-09-08
Fix routes nested under a translated sub-map reversing to the default-locale path

treagod:
Routes nested under a sub-map mounted on a translated path reverse to the default-locale path, while still receiving the active locale's prefix:

```crystal
Marten.routes.draw do
  localized(prefix_default_locale: false) do
    path t("routes.articles"), ARTICLE_ROUTES, name: "articles"   # ARTICLE_ROUTES: path "/create", …, name: "create"
  end
end

I18n.with_locale(:en) { Marten.routes.reverse("articles:create") }
# => "/en/artikel/create"   English prefix, German path — raises NoResolveMatch when resolved
```

`#reverse` and `#resolve` disagree, so every {% url %} under the sub-map emits a link that 404s in non-default locales. It fails silently and only outside the default locale.


ellmetha (approved):
Nice fix. Thank you!

PR #430, opened 2026-09-14
fix: handle multipart data without body in request

miry:
Currently, if the GET HTTP request has `Content-Type: multipart/form-data` and an empty body, it raises the following error: `Cannot extract form-data from HTTP request: could not find boundary in Content-Type (HTTP::FormData::Error)`.

```
Started "GET /foo" at 2026-09-14 18:51:42 UTC...
2026-09-14T18:51:42.594302Z [http.server] Unhandled exception
Cannot extract form-data from HTTP request: could not find boundary in Content-Type (HTTP::FormData::Error)
  from /usr/share/crystal/src/http/formdata.cr:123:5 in 'extract_raw_data_params'
  from lib/marten/src/marten/http/request.cr:65:36 in 'data'
  from lib/marten/src/marten/server/handlers/debug_logger.cr:13:15 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:30:7 in 'call_next'
  from /usr/share/crystal/src/http/server/handlers/error_handler.cr:18:5 in 'call'
  from /usr/share/crystal/src/http/server/request_processor.cr:58:11 in 'process'
  from /usr/share/crystal/src/http/server.cr:521:5 in 'handle_client'
  from /usr/share/crystal/src/http/server.cr:451:5 in '->'
  from /usr/share/crystal/src/fiber.cr:168:11 in 'run'
  from /usr/share/crystal/src/fiber.cr:103:3 in '->'
  from ???
```

## Proposal

To avoid this, don't try to parse the body and return an empty result instead.


ellmetha (approved):
Thanks for the fix!

